PR #31, opened 2015-01-13
Add support for detached Ed25519 signatures

kmcallister:


dnaq:
Thanks for the pull request. Unfortunately I won't be able to merge it since your api differs from the other API's that return detached tags (like auth and onetimeauth).

If you can implement a similar newtype-based API then I'm open to merging, but I prefer to not need dynamic memory allocation for fixed-size values.

So something like

```
pub struct Signature(pub [u8; SIGNATUREBYTES]);
impl Eq for Signature {}
impl PartialEq for Signature {
    // safe comparison function
}
newtype_clone!(Signature);
newtype_impl!(Signature, SIGNATUREBYTES);
```

Issue #21, opened 2014-12-01
scrypt support


tarcieri:
It'd be nice if the `crypto_pwhash_scryptsalsa208sha256` and `crypto_pwhash_scryptsalsa208sha256_ll` APIs were exposed.

This is something I can work on myself if you'd like to see it added.

dnaq:
Sorry that it has taken so long, I lost a lot of work due to hardware failure.

tarcieri:
Me too, been meaning to work on this myself, but if you'd like to go for it, be my guest ;)

I saw that there's an implementation in rust-crypto as well (which confusingly shares the same "crypto" namespace)

tarcieri:
@dnaq sweet!

dnaq:
@tarcieri please take a look and see if you're happy with the exposed API. I'm open to changing it if needed.

tarcieri:
Definitely good enough to get me started, thanks!

PR #34, opened 2015-01-21
libsodium-sys: Added bindings from crypto_pwhash_scryptsalsa208sha256.h

actimehl:
First step for Issue #21

### DIFF
--- a/libsodium-sys/lib.rs
+++ b/libsodium-sys/lib.rs
@@ -2,7 +2,7 @@
 #![allow(unstable)]
 
 extern crate libc;
-use libc::{c_int, c_ulonglong, c_char, c_uchar, size_t, uint8_t, uint32_t, uint64_t};
+use libc::{c_int, c_ulonglong, c_char, c_uchar, size_t};
 
 
 // aead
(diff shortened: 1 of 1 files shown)


dnaq:
Thanks, I just have two comments.
1. Why do you import c_uintXX? rusts uXX types are already of known and fixed size. (e.g. c_uint32 -> u32)
2. Some functions have pointers to arrays as arguments, however it would probably be better for the API that these functions are exposed as straight pointers, otherwise we need to do pointer casting to call them with for example slices.

dnaq:
Will you update the pull-request or should I close it?

actimehl:
I'll update

actimehl:
Changed c types to rust ones if its unambiguous.
And it's now impossible to see, that crypto_pwhash_scryptsalsa208sha256_str expects a fix size array and not a null terminated c-string.

dnaq:
Thanks. u8, u16, et. al are well defined in the rust documentation, so I'm not worried about that at all. In regards to the fixed size array, if you're sure that rust compiles '*mut u8' to the same thing as '*mut [u8; SIZE]' (and even better, show where that is documented) then I might change all ffi-functions that take fixed length arguments (and replace all calls to 'as_ptr()' with as '*[u8; SIZE]').

actimehl:
In c `void function(char *data);` and `void function(char data[10]);` are the same. But the second declaration tells what the function expects. In c one can call `function` in both cases with any array or pointer and the compiler won't bother.

But rust will take the hint and stop you from calling `function` with an array of wrong size. Look at [rust playpen](http://is.gd/ePibAZ) and compile to asm to see, that both calls are compiled to the same asm. See also https://github.com/actimehl/c_ffi_test

dnaq:
I updated all other code in libsoidum-sys to take fixed-length arrays where possible as well.

PR #38, opened 2015-02-11
libsodium-sys: split lib.rs in multiple files

actimehl:
Splits lib.rs in multiple files corresponding to the original c header files. Now it's easier to maintain libsoidum-sys.


dnaq:
Thanks for this pull request, it'll make the maintenance of libsodium-sys a lot easier in the future. It'll take a while for me to review the patch though (nothing personal, I review all commits to sodium oxide since it's a cryptographic library), that's why I haven't merged it yet.

Just wanted to let you know that the patch is well appreciated.

PR #42, opened 2015-03-09
replace deprecared push_all with extend

actimehl:


dnaq:
Where do you find a deprecation warning for `push_all()`. rustc 2015-03-14 does not complain about deprecation, and it's not showed as deprecated in the standard library documentation.

actimehl:
You are right, I got this warning after removing `#![feature(collections)]`

```
src/utils.rs:14:9: 14:22 warning: use of unstable library feature 'collections': likely to be replaced by a more optimized extend
src/utils.rs:14     dst.push_all(buf);
                        ^~~~~~~~~~~~~
src/utils.rs:14:22: 14:22 help: add #![feature(collections)] to the crate attributes to silence this warning
```

actimehl:
Now the pull request makes sense

dnaq:
As of now when benchmarking the performance still seems to be slightly better when using `push_all()`.

`push_all()`
test crypto::secretbox::xsalsa20poly1305::bench::bench_seal_open                ... bench:    194737 ns/iter (+/- 6228)

`extend()`
test crypto::secretbox::xsalsa20poly1305::bench::bench_seal_open                ... bench:    213381 ns/iter (+/- 14673)

Because of that I think I'll hold back on merging the change until the more optimized `extend()` has landed.

PR #62, opened 2015-05-11
Implement PartialEq and Eq for crypto::sign::PublicKey

benjaminbollen:
For use in MaidSafe's Sentinel library we need to search for crypto::sign::PublicKeys in a collection.  Please consider this patch to implement PartialEq and Eq for public signing keys.

Signed-off-by: Benjamin Bollen <benjamin.bollen@maidsafe.net>


dnaq:
Done, I also did the same for all other public key types in sodiumoxide

benjaminbollen:
cheers !

PR #72, opened 2015-06-15
Zero scalars when dropped

burdges:
Afaik scalars are always sensitive information, even in uncommon uses,
so they should be erased when dropped.


burdges:
Oops, I'm a bt too hasty there.  :)

burdges:
I'll fix it and resubmit.

burdges:
Actually this builds and passes the test suite on my machine.  I'll update it if I can tell what's going wrong on travis-ci

burdges:
Oops, I'm a bit too hasty there.  I'll fix it and resubmit.

dnaq:
I think tracks has problems building rust code right now. You could take a look at the travis build log to see what the actual failure is. My guess is that rustup times out.

burdges:
It's fine now : https://github.com/dnaq/sodiumoxide/pull/73

PR #75, opened 2015-07-01
Made most types encodable and decodable.

Fraser999:
The main thrust of this PR is to make most of the data types encodable and decodable by adding the relevant `rustc_serialize` traits inside the `newtype_impl` macro.

We've also added tests for these functions to all affected types.

There are a few places where we have applied `Eq` and `PartialEq` to non-secret types.


dirvine:
This PR would be an enormous help in our code base as at the moment we do have to copy and serialise, which is not great. So this would clean a lot up. For some reason using a git TOT does not work as the libsodium-sys is not built for some reason (Even including that separately). So this PR resolves a lot of blocks and untidy code for us, if it's merged but also version bumped in crates.io. So a big please :-)

dirvine:
I should say actually, we can happily add this to our local jenkins for win/osx testing if that helps you out. It will help us anyway to make sure all libs are cross platform/OS.

burdges:
Nice :) 

I wondered if any `newtype_*` macros should be exported, not necessarily `newtype_impl`, but maybe others.  I even started a pull request that renamed `newtype_drop` to `newtype_zerodrop` and exported it, but then descided I wasn't completely sure they should be exported. 

Thoughts?  If so, we might want to export a seperate `newtype_serialize` macro too.

dirvine:
Interesting, but I cannot see why really, sorry if I am misunderstanding but for Traits on these types alone then the macro's being private is probably correct? As I say I think I am mis-understanding the point perhaps. 

I know Rust serialise will handle arrays soon enough, in this case though it may be making some of these specifically not have some traits (copy etc. arguable index as well) is probably a good thing for users. I think serialise/parse are probably required in most cases, but perhaps all the types need to extensively borrow to preserve the memory overwriting.

burdges:
Interesting.  Appears the `rustc_serialize` package supports the key sizes here since https://github.com/rust-lang/rustc-serialize/issues/55  Do  `derive(RustcEncodable, RustcDecodable)` not work on the newtypes here?  Odd.  Anyways, you clarified my thinking of why not to export them, thanks.  :)

dirvine:
No problem at all, I think the issue is these are not really arrays but types that contain only an array and are not seen by serialise as an array (they are for encode) when parsing from a stream. They encode OK but as arrays and not these types (i.e. `struct SecretKey { pub [u32; 8] }`). If they were only arrays named in a struct I think it may work. `rustc_serialize`  seems to be a bit confused with these newtypes.

burdges:
Strange.  It sounds like `rustc_serialize` cannot handle structs then?  In Haskell, it's perfectly normal to create a newtype to drop some typeclass (trait) but derive other existing traits.

Fraser999:
I think it's more down to a limit on the size of array which rustc-serialize can handle.  The handling of fixed-sized arrays is [implemented via a macro](https://github.com/rust-lang/rustc-serialize/blob/d50c015b9858cd699004af4054e013886873c971/src/serialize.rs#L572-L606) and the maximum size is 32.

We could otherwise just derive the `RustcEncodable` and `RustcDecodable` traits as you suggest.

dirvine:
Yes I initially tried a mechanism to encode all the array sizes here, in the macro, but of course collided when more than one type had the same size. so I ended up explicitly stating the type. Bit of a merry go round, but at least works and seems solid. I am sure this will improve, the existing macro decode seems a bit messy in Rust. I am sure it will improve a lot.

dnaq:
Hi. Thanks for the pull request and sorry that it's taken a while for me to answer. Thanks for taking your time to contribute to sodium oxide. I've reviewed your code and I have a couple of code review comments:

## General notes
This will incur a dependency on `rustc_serialize` that might reduce the usability of the library in more constrained environments, e.g. embedded systems.

Would it be possible for you to use slice notation and the `from_slice()` method to serialize/deserialize the types instead of adding the dependency?

## hash_macros.rs
`PartialEq` for `Hash` is defined using slice comparison, this can lead to
side channel attacks if someone uses the hash function for e.g. hashing passwords or uses it for implementing HMAC. A design goal of sodium oxide is to be misuse-resistant, i.e. to make it hard for users of the library to use it in ways that are cryptographically broken.

## scryptsalsa208sha256.rs
`PartialEq` for `HashedPassword` is defined using slice comparisons. This might lead to side channel attacks if a user (ab)uses the library and tries to verify two `HashedPassword`s with each other instead of using the `pwhash_verify()` function.

## test_utils.rs
I'm not a big fan of the `CheckEq` trait. Shouldn't all types tested have `PartialEq` and `Eq` implementations, and if so, couldn't the `CheckEq` trait be replaced by `assert(x == y)` or `assert_eq(x, y)`?

Fraser999:
Thanks for the response.

> Would it be possible for you to use slice notation and the `from_slice()` method to serialize/deserialize the types instead of adding the dependency?

I'm not quite sure which `from_slice()` method you're referring to, but we have been using the `&value[..]` technique to serialise some types (where their `len()` is less than or equal to 32) and have had to convert other types to a `Vec<u8>` since rustc_serialize doesn't handle arrays above 32.

If we have these crypto types derive from the rustc_serialize traits, we cut out an awful lot of boilerplate - we can just have our struct derive the same traits.  This seems to be not only more convenient, but also safer, since users aren't forced to reimplement encode and decode functions for each of their structs which hold a crypto type.  Implementing the encode/decode functions by hand is inevitably more error-prone than just deriving the traits I feel.

I hadn't considered the notion that some more constrained environments might not want to depend on rustc_serialize though to be honest.

Do you think we could enable the encode/decode functions by default, but disable them (and the rustc_serialize dependency) via a feature?  I haven't used features myself yet - I'm not even sure if this is possible.

Regarding your other three points (the `PartialEq` and `CheckEq` traits) - I have already got another branch ready where I have exposed libsodium's `sodium_memcmp` function and use that inside the `newtype_impl` macro to implement `PartialEq` for all the types.

I think this addresses all your remaining points.  I didn't want to create that PR until I heard from you about this one, since that branch is based off of this one.  I can append that commit to this PR if you'd like.  The commit I'm referring to is https://github.com/Fraser999/sodiumoxide/commit/e586dd60c5b28e40ba6200fa1d5306b62131b207

I was also considering adding the `Debug` trait inside the macro so that all types (even secret) can be printed - I can't see an issue with that.  Do you have an opinion on this?  Or do you prefer to discuss that somewhere other than this PR?

dnaq:
> I'm not quite sure which from_slice() method you're referring to, 

All newtypes in sodium oxide have a `from_slice()` method that creates the newtype from a byte slice (&[u8]) (by copying)

> but we have been using the &value[..] technique to serialise some types (where their len() is less than or equal to 32) and have had to convert other types to a Vec<u8> since rustc_serialize doesn't handle arrays above 32.

Do you think `rustc_serialize` would accept a patch for larger arrays?

> If we have these crypto types derive from the rustc_serialize traits, we cut out an awful lot of boilerplate - we can just have our struct derive the same traits. This seems to be not only more convenient, but also safer, since users aren't forced to reimplement encode and decode functions for each of their structs which hold a crypto type. Implementing the encode/decode functions by hand is inevitably more error-prone than just deriving the traits I feel.

This is a strong argument for your case, I'll have to think it over a bit.

> Do you think we could enable the encode/decode functions by default, but disable them (and the rustc_serialize dependency) via a feature? I haven't used features myself yet - I'm not even sure if this is possible.

This should be possible using features and that might be the best way forward.

> Regarding your other three points (the PartialEq and CheckEq traits) - I have already got another branch ready where I have exposed libsodium's sodium_memcmp function and use that inside the newtype_impl macro to implement PartialEq for all the types.

That sounds good, could you amend those changes to this pull request?

> I was also considering adding the Debug trait inside the macro so that all types (even secret) can be printed - I can't see an issue with that. Do you have an opinion on this? Or do you prefer to discuss that somewhere other than this PR?

That is ok by me, but it should be in its own pull request.

Fraser999:
> All newtypes in sodium oxide have a `from_slice()` method that creates the newtype from a byte slice (&[u8]) (by copying)

Ah - I see thanks.  Yes that could be useful.

> Do you think `rustc_serialize` would accept a patch for larger arrays?

I did consider a PR there, and if was accepted it would make this PR much simpler (and probably safer) since the affected structs could just derive the traits `RustcDecodable` and `RustcEncodable` rather than implementing these traits inside the macro.

It looks like the `scryptsalsa208sha256::HashedPassword` is the longest array in the library at 102 elements - all the others are 64 or less.  I'll see if I can get that into the rustc_serialize repo, but it's quite a bit bigger then the current limit.

> This should be possible using features and that might be the best way forward.

I think so too.  Would you need me to implement this in order to accept this PR, or would I be able to tackle this as a subsequent PR?

> That sounds good, could you amend those changes to this pull request?

Done now.

> That is ok by me, but it should be in its own pull request.

Great.  I'll let the dust settle on this one first I think :)

Fraser999:
I had a look at the `rustc_serialize` code to see about increasing the size of arrays it can handle.  Unfortunately its array-handling is [implemented via a recursive macro](https://github.com/rust-lang/rustc-serialize/blob/master/src/serialize.rs#L572-L606), and we hit the recursion limit at an array of size 63.

There's already a request for larger sizes made by @dirvine in a comment in https://github.com/rust-lang/rustc-serialize/pull/117 but it's unanswered so far.

Fraser999:
I've addressed your comments as requested (all of them good points - thanks).

dnaq:
Finally merged your pull request (after a few small changes), sorry that it took so long. I'm still interested in a patch that makes rustc-serialize an optional feature if you have the time. Thanks again for the pull request.

Fraser999:
That's great thanks :)

I've submitted the new PR as discussed in this thread.  By the way, what's your plan with regards to getting these changes into crates.io?

I'll have a look at making rustc-serialize optional next.

dnaq:
I'll upload a new version to crates.io as soon as I've merged the other changes (I'm on vacation at the moment, so that's why I've only updated intermittently).

dnaq:
Just releases sodiumoxide-0.0.6

PR #76, opened 2015-07-22
Implement other traits

Fraser999:
This implements `::std::fmt::Debug` for all `newtype`s.

It also provides a new macro which implements further useful traits for non-secret types which allows them to be held in any of the `std::collections` types.


dnaq:
I removed the `non_secret_newtype_impl!` macro call from `HashedPassword`, but other than that your pull request is merged now (and released on crates.io).

Fraser999:
Thanks very much!

PR #82, opened 2015-08-17
sodium_memcmp() expects a size_t

romanb:
Since [sodium_memcmp expects a size_t as the length](https://github.com/dnaq/sodiumoxide/blob/6ab8f7cf501288237f999c7a1ef79c0931d8cc96/libsodium-sys/src/utils.rs#L5), when building on (or cross-compiling for) 32-bit architectures where `size_t` is a `u32`, enforcing a `u64` via `as u64` in `safe_memcmp` causes the code to not compile.


dnaq:
Thanks. Good catch.

Issue #85, opened 2015-09-02
Missing Ed25519 to Curve25519 key conversion


istankovic:
It would be nice to have bindings for these:
```
crypto_sign_ed25519_pk_to_curve25519
crypto_sign_ed25519_sk_to_curve25519
```

dnaq:
I've been holding off adding these functions since using the same key for different purposes carries the risk of opening you up to cross protocol attacks unless you're really careful. I might add them, but with a scary warning in the documentation, but I'd prefer not to to be honest.

What is your use case?

istankovic:
While I basically agree, I have to point out that if one is not careful with nonces,
one may very well end up with a broken system (even more easily, I'd say, than by
reusing the same key for signing and encryption).  And as far as I can see, sodiumoxide
does have bindings for functions taking nonces...

dnaq:
I fully agree.The problem is that (with the primitives in NaCl) nonces are a necessity for creating a working cryptosystem. Reusing the same key can usually be avoided (e.g. by using key derivation from a master key).

There are misuse-resistance block cipher modes, where nonce-reuse isn't that big of a problem, e.g. SIV-mode, but NaCl doesn't expose any misuse-resistant primitives.

That being said I will probably add the bindings to the two functions you requested, but with a suitable warning. I'm still interested in your use case though.

eminence:
I think my use-case is just easy-of-use.  Sometimes I want to encrypt data, other times I just want to sign it (since I believe working with signatures is faster than encrypting).    Tt would be nice for the receiving end to have to manage only 1 public key, not two

dnaq:
Signatures aren't much faster than encryption (both `crypto::box_::seal()` and `crypto::sign::sign()` do one curve multiplication). However they are for completely different use-cases. e.g. `crypto::box_` uses public key authenticators and does not give you non-repudiation whereas `crypto::sign` gives you non-repudiation.

I'm going to close this issue for now since I haven't seen a security proof that says that cross protocol attacks are impossible when using the same key for both curve25519 and ed25519.

eminence:
To play devil's advocate: if libsodium provides this function, should it also be exposed by the bindings?   It could come with a similar warning for scalarmut and friends:

> Unless you know what you're doing you most certainly don't want to use the functions in `crypto::scalarmult`, `crypto::stream`, `crypto::auth` and `crypto::onetimeauth`.


[...]

PR #89, opened 2015-11-22
Additions for keypair conversion from ed25519 to curve25519

cjcole:
http://doc.libsodium.org/advanced/ed25519-curve25519.html

### DIFF
--- a/libsodium-sys/src/crypto_sign_ed25519.rs
+++ b/libsodium-sys/src/crypto_sign_ed25519.rs
@@ -41,6 +41,12 @@ extern {
     pub fn crypto_sign_ed25519_seedbytes() -> size_t;
     pub fn crypto_sign_ed25519_publickeybytes() -> size_t;
     pub fn crypto_sign_ed25519_secretkeybytes() -> size_t;
+    pub fn crypto_sign_ed25519_pk_to_curve25519(
+        curve25519_pk: *mut [u8; crypto_box_curve25519xsalsa20poly1305_PUBLICKEYBYTES],
+        ed25519_pk: *const [u8; crypto_sign_ed25519_PUBLICKEYBYTES]) -> c_int;
+    pub fn crypto_sign_ed25519_sk_to_curve25519(
+        curve25519_sk: *mut [u8; crypto_box_curve25519xsalsa20poly1305_SECRETKEYBYTES],
+        ed25519_sk: *const [u8; crypto_sign_ed25519_SECRETKEYBYTES]) -> c_int;
 }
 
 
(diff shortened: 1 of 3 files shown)


dnaq:
Thanks for the pull request. We've had some discussion on this functionality in issue #85, and I decided to not add it until someone could present a valid use-case for the functionality. The reason being that I don't feel confident in adding something that could open up to cross-protocol attacks unless there's a real need for it from developers. Do you have a specific use-case you need?

burdges:
Pond uses these when initializing v2 of its ratchet, but Pond is not written in Rust and it didn't need them really.  I could imagine Axolotl initialization with TripleDH might tend towards using these when you're trying to squeeze a bunch of preliminary ratchet state onto a business card QR code or something.

dnaq:
I'm not familiar with the Pond ratchet, but AFAIK the Axolotl ratchet only uses diffie-hellman, e.g. `crypto_scalarmult()` and doesn't use public key signatures, `crypto_sign()` so there would be no need to convert between ed25519 and curve25519 key types for it. I might be wrong though, I don't know if the Axolotl ratchet has been updated since I last looked at it.

burdges:
Pond uses Axolotl.  Imho, it's not worth going into how Pond uses these. 

Yes, Axolotl normally uses TripleDH for deniability, but there are situations where you might want full/manual ratchet initialization, like a phones initiating connectivity in physical proximity or a business card with a QR code.  Those applications benefit from a signing key that can be verified elsewhere, so they save 32 bytes using this. 

There are applications in implicit signatures like elliptic curve qu-vanstone that save 64 bytes, but that requires exposing even more.  https://en.wikipedia.org/wiki/Implicit_certificate#Generating_implicit_certificates

dnaq:
I'm closing this issue for now since we're mostly discussing hypothetical use cases for a feature that brings with it potential for cross protocol attacks. I'm open to revisiting this issue if more concrete use-cases are found. Thanks to all of you that have brought valuable input to this issue.

cjcole:
Just curious. It is not the case that sodiumoxide is intended to be a generic wrapper for libsodium?

jedisct1:
sodium is a bit more opinionated.

If you are looking for a low-level, generic wrapper for libsodium, https://github.com/rustyhorde/sodium-sys and https://github.com/zonyitoo/libsodium-sys might be better options.

cjcole:
One of those might be a better fit for me. Thanks for the links.

dnaq:
> It is not the case that sodiumoxide is intended to be a generic wrapper for libsodium?

sodiumoxide is a bit more opinionated. Hope you find what you're looking for elsewhere. However, I asked you for a use case for the functionality you requested, but you didn't provide me one. Still if you have a specific use case that you want to discuss we could discuss it further.

PR #92, opened 2015-12-09
Increased libsodium version in Travis script to 1.0.6.

Fraser999:
When linking to libsodium 1.0.3, we get the following error:

>     undefined reference to `sodium_increment'

Version 1.0.4 links OK.

I'm not sure if you want to document 1.0.4 as a minimum requirement anywhere (e.g. the README) or even if we can check for this in build.rs?

I would have changed the dependency in the Travis script to the latest version (1.0.7) but this fails to link for me when using the prebuilt MinGW libraries.  I haven't assessed this issue yet, but my initial suspicions are that the prebuilt libs are defective since they're significantly smaller than those in 1.0.6, and libsodium's changelog doesn't indicate any significant reduction of its API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: rust
 install:
-- wget https://github.com/jedisct1/libsodium/releases/download/1.0.3/libsodium-1.0.3.tar.gz
-- tar xvfz libsodium-1.0.3.tar.gz
-- cd libsodium-1.0.3 && ./configure --prefix=/usr && make && sudo make install &&
+- wget https://github.com/jedisct1/libsodium/releases/download/1.0.6/libsodium-1.0.6.tar.gz
+- tar xvfz libsodium-1.0.6.tar.gz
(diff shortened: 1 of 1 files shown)


dnaq:
Thanks. I never got an email from travis for some reason.

Fraser999:
Thanks for merging this.

Regarding libsodium 1.0.7 for MinGW, the issue has been fixed: https://github.com/jedisct1/libsodium/issues/327.  It was intentional, but has been reverted.

However in version 1.0.7, the `edwards25519sha512batch` functions have been labelled deprecated and moved to a file named `obsolete.c`.  In doing so, the signature verification has picked up a bug which causes your test `crypto::sign::edwards25519sha512batch::test::test_sign_verify` to fail (there are no tests for this in libsodium that I can see).

I believe I've found the source of the issue in libsodium and will submit a PR to fix it shortly.  Until that's merged and the prebuilt binaries updated again, we still can't use the latest version :)

Fraser999:
PR to libsodium at https://github.com/jedisct1/libsodium/pull/329.

PR #95, opened 2015-12-10
Used call_once in init() to make it thread-safe and updated documentation.

Fraser999:
I believe this makes it safe for users to call `init()` concurrently.

The test doesn't check that any `init()` call actually succeeded, only that it didn't execute the call more than once.  Ideally I'd like to have checked that it returned success exactly once, but I don't know if that's reasonable - i.e. if there are environments where the library can still be used safely despite `init()` returning `false`.

Also, since the tests run in parallel by default, I think we should be calling `init()` at the start of every test.  This PR makes it safe to do that now, although I plan to make that change in a separate PR if you agree and are OK with this one.

### DIFF
--- a/src/crypto/auth/auth_macros.rs
+++ b/src/crypto/auth/auth_macros.rs
@@ -31,9 +31,9 @@ new_type! {
 
 /// `gen_key()` randomly generates a key for authentication
 ///
-/// THREAD SAFETY: `gen_key()` is thread-safe provided that you have
-/// called `sodiumoxide::init()` once before using any other function
-/// from sodiumoxide.
+/// THREAD SAFETY: `gen_key()` is thread-safe provided
+/// [`sodiumoxide::init()`](../../../fn.init.html) has been called and has returned
+/// `true` before using any other function from sodiumoxide.
 pub fn gen_key() -> Key {
     let mut k = [0; KEYBYTES];
     randombytes_into(&mut k);
--- a/src/crypto/box_/curve25519xsalsa20poly1305.rs
+++ b/src/crypto/box_/curve25519xsalsa20poly1305.rs
@@ -44,9 +44,9 @@ new_type! {
 
 /// `gen_keypair()` randomly generates a secret key and a corresponding public key.
 ///
-/// THREAD SAFETY: `gen_keypair()` is thread-safe provided that you have
-/// called `sodiumoxide::init()` once before using any other function
-/// from sodiumoxide.
+/// THREAD SAFETY: `gen_keypair()` is thread-safe provided
+/// [`sodiumoxide::init()`](../../../fn.init.html) has been called and has returned
+/// `true` before using any other function from sodiumoxide.
 pub fn gen_keypair() -> (PublicKey, SecretKey) {
     unsafe {
         let mut pk = [0u8; PUBLICKEYBYTES];
@@ -60,9 +60,9 @@ pub fn gen_keypair() -> (PublicKey, SecretKey) {
 
 /// `gen_nonce()` randomly generates a nonce
 ///
-/// THREAD SAFETY: `gen_nonce()` is thread-safe provided that you have
-/// called `sodiumoxide::init()` once before using any other function
-/// from sodiumoxide.
+/// THREAD SAFETY: `gen_nonce()` is thread-safe provided
+/// [`sodiumoxide::init()`](../../../fn.init.html) has been called and has returned
+/// `true` before using any other function from sodiumoxide.
 pub fn gen_nonce() -> Nonce {
     let mut n = [0; NONCEBYTES];
     randombytes_into(&mut n);
--- a/src/crypto/pwhash/scryptsalsa208sha256.rs
+++ b/src/crypto/pwhash/scryptsalsa208sha256.rs
@@ -62,8 +62,9 @@ new_type! {
 
 /// `gen_salt()` randombly generates a new `Salt` for key derivation
 ///
-/// THREAD SAFETY: `gen_salt()` is thread-safe provided that you have called
-/// `sodiumoxide::init()` once before using any other function from sodiumoxide.
+/// THREAD SAFETY: `gen_salt()` is thread-safe provided
+/// [`sodiumoxide::init()`](../../../fn.init.html) has been called and has returned
+/// `true` before using any other function from sodiumoxide.
 pub fn gen_salt() -> Salt {
     let mut salt = Salt([0; SALTBYTES]);
     {
--- a/src/crypto/secretbox/xsalsa20poly1305.rs
+++ b/src/crypto/secretbox/xsalsa20poly1305.rs
@@ -33,9 +33,9 @@ const BOXZEROBYTES: usize = 16;
 
 /// `gen_key()` randomly generates a secret key
 ///
-/// THREAD SAFETY: `gen_key()` is thread-safe provided that you have
-/// called `sodiumoxide::init()` once before using any other function
-/// from sodiumoxide.
+/// THREAD SAFETY: `gen_key()` is thread-safe provided
+/// [`sodiumoxide::init()`](../../../fn.init.html) has been called and has returned
+/// `true` before using any other function from sodiumoxide.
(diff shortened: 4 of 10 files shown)


dnaq:
I've been taking a look at this pull request and while I agree that it's desirable to be able to call `init()` multiple times I'm not as sure about the semantics of `init()` returning `false` after the first time. We wouldn't want callers to start ignoring the return value of `sodium_init()`. The cleanest solution would be to use a `StaticMutex` and allowing multiple calls to `sodium_init()`, but `StaticMutex` is still feature-gated.

Another possible solution is to use `Once` and to `panic!()` on failed library initialization, but I'm of the opinion that libraries shouldn't `panic!()` unless absolutely necessary, and a failed call to `init()` is a recoverable error (even though it's not recoverable when using `Once`).

Is it ok with you if we hold back on this change until `StaticMutex` or something similar lands?

Fraser999:
> I'm not as sure about the semantics of init() returning false after the first time.

I was trying to match the behaviour of [libsodium's `sodium_init`](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/sodium/core.c), which I think we're not currently doing.  Just now, we're returning `true` unconditionally from `init`, since `sodium_init` will never return `-1`.

Assuming a single threaded environment, `sodium_init` will unconditionally return `0` the first time it runs, and every call thereafter will unconditionally return `1`.  In a multi-threaded environment, I believe we have undefined behaviour.

Since `sodium_init` doesn't return any failure except to indicate that it has been previously called, if we want to return `true` from `init` *after* it's already been called, then that means we're always returning `true`, in which case it should just return `()` I suppose?

I'm not sure there's any value in returning a bool - I can't see how it would be useful to a user, but that does seem to be the closest match to libsodium's behaviour.

dnaq:
You are completely correct in that the actual code of libsodium never returns `-1`. However [the documentation](http://doc.libsodium.org/usage/index.html) says that `-1` can be returned when initialization fails. @jedisct1, do you have any plans to add initialization that can fail to `sodium_init()` or is it safe to assume that `sodium_init()` will never return `-1` in the future?

If it is safe the easiest thing to do would be to let `init()` use `ONCE` and to not return anything from `init`.

jedisct1:
It is not safe to assume that it won't return `-1`, and the function was marked `warn_unused_result` for this reason.
You have to check that it doesn't return `-1`.

dnaq:
@jedisct1 thanks for your reply. @Fraser999 I guess we'll hold off on implementing this change until static mutexes arrive in rust stable. Closing for now.

Fraser999:
@jedisct1 Could you just clarify the planned behaviour then please?  Specifically for the case where `sodium_init()` is called multiple times.

I'm guessing that the first time it will return `0` or `-1` and every attempt thereafter will return `1` regardless of whether the initialisation succeeded or failed?  i.e. if it fails on the first attempt, there would be no point in attempting the init again.

jedisct1:
If a previous call returned `-1`, the only state it can transition to is `0`.
If a previous call returned `0`, the only state it can transition to is `1`

Fraser999:
@jedisct1 OK - thanks.

betafive:
I've looked at this as I need a thread-safe way to call sodiumoxide::init. I'm fairly new to rust but to my understanding it should be possible to use call_once and still capture the result of the sodium_init method into a bool outside of the call_once closure. I've thrown together some code which looks like this:

```
pub fn init() -> bool {
    let mut retval = true;

    INIT_ONCE.call_once(|| {
        unsafe {
            retval = ffi::sodium_init() != -1;
        }
    });

    retval
}
```

This returns true unless sodium_init() is called and fails. So sodiumoxide::init can be called multiple times and you just need to check that each call returns true.

Is this a desirable behaviour for sodiumoxide::init? If so I can submit a merge request.

We can always replace this with code using StaticMutex at a later date, but I don't think we should leave this unresolved until then (unless StaticMutex is due in stable very soon).

dnaq:
@betafive The issue with your code is that if the first call to `init()` fails (returns `false`) then all subsequent calls will still return `true`, even though `sodium_init()` isn't called anymore. This will be really confusing for users of sodiumoxide and there will be no way to recover from a failed `init()` call.

I'm still going to hold off on fixing this until `StaticMutex` lands. In the meantime might I suggest using a wrapper function that panics instead of calling `init()` directly. e.g.

```rust
pub fn init_wrapper() {
    INIT_ONCE.call_once(|| {
        let ret = sodiumoxide::init();
        assert!(ret);
    });
}
```

At the moment libsodium will never return `-1` and your function will never panic, but this will change in the future according to the author.

betafive:
That makes sense. Not as simple as I thought!

StaticMutex will make this a lot easier to get right. It probably is best to wait until this lands in rust stable and then solve it that way.

stouset:
This seems relatively easy to resolve, no?

```rust
static INIT:        Once = ONCE_INIT;
static initialized: bool = false;

pub fn init() -> bool {
    INIT.call_once(|| {
        initialized = (unsafe { ffi::sodium_init() } >= 0);
    });

    initialized
}

dnaq:
@stouset That looks like a data-race.

stouset:
How so? From the [docs](https://doc.rust-lang.org/std/sync/struct.Once.html):

> This method will block the calling thread if another initialization routine is currently running.

`initialized` can only be written to by one thread at a time, and only once. From then it may be read from multiple threads simultaneously.

dnaq:
See the edit to my comment

stouset:
Do you still feel that it's a data race?

I agree that there won't be any way to retry. But I think that's an extraordinarily insignificant use-case: libsodium doesn't even implement failure currently, and if it did, the number of retry-able failure cases seems limited. On the other hand, the current approach is clearly not thread-safe, which makes it impossible to safely write tests for anything dependent upon initialization without writing basically the same code I have pasted here.

The win is large, and well worth the extremely minor downside. When and if StaticMutex is ever stabilized, we can always replace the logic to allow for retries.

dnaq:
You're (probably) right, it isn't a data race. However I still feel vary of a library where there's no recovery path after failed initialization. I'm closing this for now, but I will revisit this issue as soon as `StaticMutex` lands.

stouset:
The API itself clearly supports retrying, even though nothing different actually happens. But realistically, there are currently zero failure modes for the API. In the event that there *are* ever failure modes implemented, the idea that simply looping over `sodium_init()` will ever succeed after it fails once seems somewhat far-fetched.

On the other hand, the lack of thread-safe initialization actively prevents tests from being correct. From the docs, "`sodium_init()` initializes the library and should be called before any other function provided by Sodium". `sodium_init()` is currently being called zero times in tests (despite functions like `randombytes` being called, which explicitly depend on initialization).

Should you implement and test FFI for functions like `sodium_malloc()`, this change will actively become necessary as a) the functions in that family very much require initialization, and b) tests run in parallel.

dnaq:
I agree that the situation is a bit sucky as is. I've also read the source code of libsodium and I know that `sodium_init` can't fail at the moment, but the author has said that this will change, and I'm a bit reluctant to assume that it won't and have the library break when/if libsodium gets updated.

I think the best way for now might be to use `Once` and to call `panic` on failed initialization (which won't happen with the current version of libsodium). Then we can switch that over to using `StaticMutex` when it lands, but we won't be caught off guard if `sodium_init` starts failing all of a sudden.

I've been reluctant to do this since I believe libraries shouldn't need to call `panic`, but I think it might be the best option for this situation.

stouset:
There's no need to call `panic`. We can simply return `true` or `false` and let the caller decide if that's panic-worthy or not. Just because repeated calls to `sodiumoxide::init()` won't change their outcome doesn't mean we can't let callers decide how to treat that failure.

dnaq:
I've mulled over your argument for a while and I think I'm convinced by the pros of thread-safety outweighing the cons of not being able to recover.

stouset:
Great! Feel free to [steal my implementation wholesale](https://github.com/stouset/secrets/blob/master/src/sodium.rs#L30).

PR #103, opened 2016-01-25
libsodium-sys: Improve generichash bindings

betafive:
These commits add additional bindings and test cases for generichash functions.

### DIFF
--- a/libsodium-sys/src/crypto_generichash.rs
+++ b/libsodium-sys/src/crypto_generichash.rs
@@ -94,3 +94,55 @@ fn test_crypto_generichash_primitive() {
         assert_eq!(s, crypto_generichash_PRIMITIVE.as_bytes());
     }
 }
+
+#[test]
+fn test_crypto_generichash_statebytes() {
+    assert!(unsafe { crypto_generichash_statebytes() } > 0);
+}
+
+#[test]
+fn test_crypto_generichash() {
+    let mut out = [0u8; crypto_generichash_BYTES];
+    let m = [0u8; 64];
+    let key = [0u8; crypto_generichash_KEYBYTES];
+
+    assert_eq!(unsafe {
+        crypto_generichash(out.as_mut_ptr(), out.len(),
+                           m.as_ptr(), m.len() as u64,
+                           key.as_ptr(), key.len())
+    }, 0);
+}
+
+#[cfg(test)]
+use std::mem;
+
+#[test]
+fn test_crypto_generichash_multipart() {
+    let mut out = [0u8; crypto_generichash_BYTES];
+    let m = [0u8; 64];
+    let key = [0u8; crypto_generichash_KEYBYTES];
+
+    let mut st = vec![0u8; (unsafe { crypto_generichash_statebytes() })];
+    let pst = unsafe { mem::transmute::<*mut u8, *mut crypto_generichash_state>(st.as_mut_ptr()) };
+
+    assert_eq!(unsafe {
+        crypto_generichash_init(pst,
+                                key.as_ptr(), key.len(),
+                                out.len())
+    }, 0);
+
+    assert_eq!(unsafe {
+        crypto_generichash_update(pst,
+                                m.as_ptr(), m.len() as u64)
+    }, 0);
+
+    assert_eq!(unsafe {
+        crypto_generichash_update(pst,
+                                m.as_ptr(), m.len() as u64)
+    }, 0);
+
+    assert_eq!(unsafe {
+        crypto_generichash_final(pst,
+                                out.as_mut_ptr(), out.len())
+    }, 0);
+}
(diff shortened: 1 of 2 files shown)


betafive:
Really not sure what's up with the check failure - it looks like it's died installing rust and not built my code.

libsodium-sys/src/crypto_generichash.rs
```diff
@@ -8,6 +8,9 @@ pub const crypto_generichash_KEYBYTES_MAX: usize = crypto_generichash_blake2b_KE
 pub const crypto_generichash_KEYBYTES: usize = crypto_generichash_blake2b_KEYBYTES;
 pub const crypto_generichash_PRIMITIVE: &'static str = "blake2b";
 
+#[allow(non_camel_case_types)]
+pub enum crypto_generichash_state { }
+
```
dnaq: I think we probably need to add the real struct definition here, please take a look at `crypto_hash_sha256.rs` to see how that's done...

libsodium-sys/src/crypto_generichash.rs
```diff
@@ -25,6 +28,27 @@ extern {
...
+        inlen: c_ulonglong)
+        -> c_int;
+
+    pub fn crypto_generichash_final(
+        state: *mut crypto_generichash_state,
+        out: *mut u8,
+        outlen: size_t)
+        -> c_int;
+
+    pub fn crypto_generichash_statebytes() -> size_t;
```
dnaq: Also add a constant `CRYPTO_GENERICHASH_STATEBYTES` and a test to see that it's equal to the value this function returns.

libsodium-sys/src/crypto_generichash.rs
```diff
@@ -70,3 +94,55 @@ fn test_crypto_generichash_primitive() {
         assert_eq!(s, crypto_generichash_PRIMITIVE.as_bytes());
     }
 }
+
+#[test]
+fn test_crypto_generichash_statebytes() {
+    assert!(unsafe { crypto_generichash_statebytes() } > 0);
```
dnaq: Don't test that it's greater than 0, test that it's equal to the known constant.

libsodium-sys/src/crypto_generichash.rs
```diff
@@ -70,3 +94,55 @@ fn test_crypto_generichash_primitive() {
...
+}
+
+#[test]
```
dnaq: The tests in libsodium-sys shouldn't be tests of the actual primitives, instead it should be tests of the bindings, e.g. that all constants are correct and so on. All test vectors of actual primitives should be defined in `crypto::generichash`. I would like to see (at least) all test vectors from `libsodium`.

dnaq:
Hi and thanks for the pull request. I wrote a bunch of comments on your code, the same comments apply to both `crypto_generichash.rs` and `crypto_generichash_blake2b.rs`.

libsodium-sys/src/crypto_generichash.rs
betafive: The libsodium docs say "The crypto_generichash_state structure is packed and its length is either 357 or 361 bytes.", I'm not sure how to represent that in rust.

The C code in question is:

```
#pragma pack(push, 1)
typedef CRYPTO_ALIGN(64) struct crypto_generichash_blake2b_state {
    uint64_t h[8];
    uint64_t t[2];
    uint64_t f[2];
    uint8_t  buf[2 * 128];
    size_t   buflen;
    uint8_t  last_node;
} crypto_generichash_blake2b_state;
#pragma pack(pop)
```

libsodium-sys/src/crypto_generichash.rs
betafive: libsodium does not have a constant for this, it calculates `(sizeof(crypto_generichash_state) + (size_t) 63U) & ~(size_t) 63U`.

libsodium-sys/src/crypto_generichash.rs
betafive: Test vectors should definitely be added at the higher level. What I was trying to address here is that the code will compile even if the libsodium function names are misspelled. By having a simple test like this that the binding can be called, we can ensure that the functions are named correctly and do exist in libsodium.

dnaq:
I need you to agree with the licensing changes in #102 before merging, is that ok with you?

betafive:
That's fine by me, I'm happy for these contributions to be under any open source license.

adamhjk:
Is there anything left to do to get this in? We would love to be using the generic hash functions.

dnaq:
I will merge it, I've been a bit pressed for time lately.

adamhjk:
Thank you so much! We're going to be using the library pretty extensively -
let me know if you need folks to help you with maintenance.

betafive:
Is there anything I can do to help get these changes merged?

dnaq:
I'm planning on starting work on a new release after the 15th of May, then
I'll be merging all of the outstanding pull requests. After that date
releases are going to become more regular again.

In the long run we might look into adding more maintainers, but I'll need
to find someone who is as paranoid and as opinionated as I am when it comes
to the design and implementation of cryptographic libraries... :)

dnaq:
Merged it, sorry for the wait...

PR #106, opened 2016-01-27
Functions that directly change memory are unsafe

stouset:
These two methods explicitly break Rust's safety guarantees. Arbitrary memory cannot be safely zeroed out or incremented as if it were an unsigned integer.


stouset:
On second glance, unlike my wrappers, these are implemented on `&[u8]` which should be safe.

PR #117, opened 2016-02-23
Loosen version constraints on dependencies

untitaker:
pkg-config 0.3.7 is out. Since openssl-sys depends on `pkg-config =
"0.3"`, Cargo selects 0.3.7. However, since libsodium-sys depends on the
exact PATCH version, we get a version conflict.

We drop the PATCH segment from our version requirements, allowing Cargo
to choose any version 0.3.y for any value of y.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,12 +10,12 @@ repository = "https://github.com/dnaq/sodiumoxide"
 version = "0.0.9"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


untitaker:
For some reason this actually is not necessary (the version conflict was due to something else), but I guess it's still nice to avoid too specific version numbers?

PR #129, opened 2016-05-19
libsodium version function

danlarkin:
This PR adds a new function, `version::version_string` - it just calls through to libsodium's `sodium_version_string()` \[1\].

This PR suits my needs, but I'm not sure it's complete:
1) Is there enough documentation?
2) Is there enough testing? I'm not sure how to do any better, but it does feel light.
3) Should I expose `sodium_library_version_major()` and `sodium_library_version_minor()`?

Thanks for your time and consideration.

\[1\]: https://github.com/jedisct1/libsodium/blob/master/src/libsodium/sodium/version.c

### DIFF
--- a/libsodium-sys/lib.rs
+++ b/libsodium-sys/lib.rs
@@ -53,3 +53,4 @@ include!("src/crypto_verify_64.rs");
 
 include!("src/randombytes.rs");
 include!("src/utils.rs");
+include!("src/version.rs");
--- a/libsodium-sys/src/version.rs
+++ b/libsodium-sys/src/version.rs
@@ -2,4 +2,6 @@
 
 extern {
     pub fn sodium_version_string() -> *const c_char;
+    pub fn sodium_library_version_major() -> c_int;
+    pub fn sodium_library_version_minor() -> c_int;
 }
--- a/libsodium-sys/src/version.rs
+++ b/libsodium-sys/src/version.rs
@@ -1,0 +1,5 @@
+// version.h
+
+extern {
+    pub fn sodium_version_string() -> *const c_char;
+}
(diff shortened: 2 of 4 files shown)


dnaq:
Thanks for the pull request. Two quick comments.

1. Could you return a &static str instead of heap allocating a String?
2. Could you add version_minor/major bindings as well? Those would be easier to use for someone needing to check the sodium version programmatically. 

Thanks

danlarkin:
Thanks for your feedback. I made the changes you mentioned, but I have some new questions:

1) wrt `version_string()`, it's calling `to_str().unwrap()` on a `CStr`. Should that be a decision left to the caller? Maybe they can't accept a possible panic? Should this function instead return `&[u8]`? OTOH, returning `&' static str` is more friendly than `&[u8]` and what are the chances that libsodium's version string contains a byte sequence incapable of utf-8 encoding?

2) `_major()` and `_minor()` both return `int`, but there's no ISA-independent way to represent this in rust. I went with `i32` as a best-effort. Is this right?

3) While they both return `int`, there are tests \[1\] to make sure they are positive. So should we map to the intent, or to the actual data type?

4) I didn't give them any tests because I don't know what I could test for... the same things libsodium already does \[1\]?

\[1\]: https://github.com/jedisct1/libsodium/blob/master/test/default/sodium_version.c#L8-L9

dnaq:
Hi again,

1. The function should return a `&'static str`, not a `&'static [u8]`. The test case should be enough to make sure that we don't get a panic when unwrapping.
2. `i32` works fine, but I've been using `usize` for unsigned constants in sodiumoxide.
3. I think it's better to map to the intent in the high-level API. The low level API should map to the exact type, e.g. `c_int` (and contain tests that the return value isn't negative). The high-level API should probably return an `usize`.
4. Yes, do those tests in the low-level API so that we are sure that it's ok to use an `usize` for the return value.

Also I need you to agree to the licensing in #102 before being able to pull your code.

danlarkin:
Okay, I feel good about this now. What do you think?

PR #130, opened 2016-06-08
Support for the libsodium aeads

JohnHeitmann:
Implement formal support for aes256gcm and chacha20poly1305. Combined modes only.

I was tempted to model their commonality in the type system, but it looks like 1) it's not project style, and 2) the commonalities start to fall apart when you look at modes outside combined mode. I can still go that route if you give me a nudge.


pijul:
Hi,

I've reviewed this pull request (I was also interested in the feature). Overall, it's a good job, but my main objection is that it allocates memory. I can see two problems with this:

- Symmetric cryptography is usually used for high performance. Allocations defeat that purpose (think of the case of a server serving tons of requests).

- As a user of this library, I would like to be able to call sodiumoxide::utils::memzero on any memory written or read by the library (even non-secrets, chacha20 uses the same functions to encrypt and decrypt). However, you're making a strong assumption on the Vec module, whose implementation might change in the future: what if secrets get copied around?

How about making sure that the caller sent a large enough buffer, instead of allocating a new one?

JohnHeitmann:
Yeah, I preferred idiomatic code over performance. The secretbox implementation made the same choice. I don't feel strongly about it, though. I'll take a pass at a zero-alloc version.

JohnHeitmann:
Updated with in-place support in addition to the original allocating version. Since you can't borrow from the same slice mutably and immutably I had to use one in_out parameter, and ask the caller for a manually-provided length.

JohnHeitmann:
Also, I tried an alternate version that returned a ``&'a [u8]`` view instead of ``usize``, like:

```rust
pub fn encrypt_in_place<'a>(in_out: &'a mut [u8], ..) -> Result<&'a [u8], ()> { .. }
```
The problem for callers is that it locks out in_out from subsequent adjacent borrows since the lifetime of the result kept the initial mutable borrow alive. This was a pain in test code; I assume it'd be a pain in real code as well.

pijul:
Hi!

Sounds cool!
In my real-world code (I'm writing an SSH server), I use mostly libsodium-sys, because I want to know where secrets are in memory.

The problem with allocating vectors is when it's in a loop, like when decrypting many packets in a loop (it's what servers and clients do, right?).

Another option to avoid this is to take a ```&mut Vec<u8>``` argument, that can reallocate. You're function returning &'a[u8] is perfectly acceptable, too.

(even though in my use case, () error type doesn't work. Either panic or return a useful type. If there's a single cause of error, known to the caller, return Option<&'a[u8]>).

As a side note, I don't use Vec for my buffers, as I want to securely erase their memory upon realloc and clear. I've written my own ```Cryptobuf``` type, with an interface (in particular bound checking) similar to ```Vec```.

I'm sorry to be so picky, crpyto code has to inspire a lot of confidence, even if it's at the price of being non-idiomatic :(

libsodium-sys/src/crypto_aead_aes256gcm.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn crypto_aead_aes256gcm_npubbytes() -> size_t;
+    pub fn crypto_aead_aes256gcm_abytes() -> size_t;
+    pub fn crypto_aead_aes256gcm_encrypt(
+        c: *mut u8,
+        clen: *mut c_ulonglong,
+        m: *const u8,
+        mlen: c_ulonglong,
+        ad: *const u8,
+        adlen: c_ulonglong,
+        nsec: *const [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: libsodium doesn't define `nsec` as const. I agree that it should be, but we need to follow the actual API so we don't get any surprises if libsodium decides to start mutating `nsec`.

libsodium-sys/src/crypto_aead_aes256gcm.rs
```diff
@@ -0,0 +1,77 @@
...
+        npub: *const [u8; crypto_aead_aes256gcm_NPUBBYTES],
+        k: *const [u8; crypto_aead_aes256gcm_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_aes256gcm_decrypt(
+        m: *mut u8,
+        mlen: *mut c_ulonglong,
+        nsec: *const [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: libsodium doesn't define `nsec` as const. I agree that it should be, but we need to follow the actual API so we don't get any surprises if libsodium decides to start mutating `nsec`.

libsodium-sys/src/crypto_aead_aes256gcm.rs
```diff
@@ -0,0 +1,77 @@
...
+        k: *const [u8; crypto_aead_aes256gcm_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_aes256gcm_encrypt_detached(
+        c: *mut u8,
+        mac: *mut u8,
+        maclen: *mut c_ulonglong,
+        m: *const u8,
+        mlen: c_ulonglong,
+        ad: *const u8,
+        adlen: c_ulonglong,
+        nsec: *const [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: libsodium doesn't define `nsec` as const. I agree that it should be, but we need to follow the actual API so we don't get any surprises if libsodium decides to start mutating `nsec`.

libsodium-sys/src/crypto_aead_aes256gcm.rs
```diff
@@ -0,0 +1,77 @@
...
+        npub: *const [u8; crypto_aead_aes256gcm_NPUBBYTES],
+        k: *const [u8; crypto_aead_aes256gcm_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_aes256gcm_decrypt_detached(
+        m: *mut u8,
+        nsec: *const [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: libsodium doesn't define `nsec` as const. I agree that it should be, but we need to follow the actual API so we don't get any surprises if libsodium decides to start mutating `nsec`.

libsodium-sys/src/crypto_aead_chacha20poly1305.rs
```diff
@@ -24,13 +24,34 @@ extern {
     pub fn crypto_aead_chacha20poly1305_decrypt(
         m: *mut u8,
         mlen: *mut c_ulonglong,
-        nsec: *mut [u8; crypto_aead_chacha20poly1305_NSECBYTES],
+        nsec: *const [u8; crypto_aead_chacha20poly1305_NSECBYTES],
```
dnaq: libsodium doesn't define `nsec` as const. I agree that it should be, but we need to follow the actual API so we don't get any surprises if libsodium decides to start mutating `nsec`.

libsodium-sys/src/crypto_aead_chacha20poly1305.rs
```diff
@@ -24,13 +24,34 @@ extern {
...
         k: *const [u8; crypto_aead_chacha20poly1305_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_chacha20poly1305_encrypt_detached(
+        c: *mut u8,
+        mac: *mut u8,
+        maclen: *mut c_ulonglong,
+        m: *const u8,
+        mlen: c_ulonglong,
+        ad: *const u8,
+        adlen: c_ulonglong,
+        nsec: *const [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: libsodium doesn't define `nsec` as const. I agree that it should be, but we need to follow the actual API so we don't get any surprises if libsodium decides to start mutating `nsec`.

src/crypto/aead/aead_macros.rs
```diff
@@ -0,0 +1,185 @@
...
+/// using a secret key `k`, and a public nonce `n`, and overwrites
+/// `in_out` with a combined byte stream of ciphertext, and an
+/// authentication tag authenticating both the plaintext and
+/// the optional additional data in `ad` .
+///
+/// The public nonce should never be re-used with the same key.
+///
+/// The encrypted length will be at most m_len + MACBYTES, so
+/// in_out.len() should be at least m_len + MACBYTES.
+pub fn encrypt_in_place(in_out: &mut [u8],
```
dnaq: The naming convention used in the library for in_place functions is `primitive_inplace`, not `primitive_in_place`

src/crypto/aead/aead_macros.rs
```diff
@@ -0,0 +1,185 @@
...
+/// `encrypt()` encrypts the plaintext `in_out` of length `m_len`
+/// using a secret key `k`, and a public nonce `n`, and overwrites
+/// `in_out` with a combined byte stream of ciphertext, and an
+/// authentication tag authenticating both the plaintext and
+/// the optional additional data in `ad` .
+///
+/// The public nonce should never be re-used with the same key.
+///
+/// The encrypted length will be at most m_len + MACBYTES, so
+/// in_out.len() should be at least m_len + MACBYTES.
```
dnaq: I'm not a fan of the API of this function. I think the best bet for the `inplace()` functions (and we should do the same for `crypto::secretbox` and `crypto::box` as well is to use the detached-functions and make the API look something like:

`encrypt_inplace(m: &mut [u8], ad: &[u8], &Nonce(ref n): &Nonce, &Key(ref k): &Key) -> Result<Tag, ()>` where `Tag` is a detached authentication tag (which we always know the size of). Of course with similar changes to `decrypt_inplace()`.

We probably also need helper functions to separate the authentication tag from an authenticated encrypted message, something like `split_tag(&'a [u8]) -> Result<(&'a Tag, &'a [u8])> to be able to easily switch between the `inplace` and the allocating versions. The name `split_tag()` isn't good though.

dnaq:
Thanks for the pull request. I have some comment on the code, but I would also like you to split the code into two different pull requests if possible. The first one should only contain the versions of the primitives that don't do inplace encryption/decryption and then we could add non-allocating primitives to both `crypto::aead`, `crypto::box` and `crypto::secretbox` at a later stage (the API should be as similar as possible). Is that ok with you?

I think we might also want to rewrite the allocating versions of `crypto::box` and `crypto::secretbox` to use the `_easy()` versions of the primitives, this should yield some performance benifit (but of course we should benchmark before switching). Maybe I should add an issue for that.

Also we should look at hiding all `sodiumoxide` primitives that require memory allocation behind a default-feature to simplify usage of `sodiumoxide` in embedded systems.

JohnHeitmann:
I removed the inplace wrappers from this PR and made nsecbytes mut again.

I agree that inplace API had was ugly. I'll play with your detached ``Result<Tag, ()>`` idea. It'll incur a 16 byte memcpy overhead for folks wanting re-combining, but it's the best compromise between safety, speed, and ergonomics I've seen so far.

JohnHeitmann:
Just wanted to check in and see if you were waiting on me. I was planning on tackling the second PR with the inplace APIs some time after this one was accepted, but if you'd like to see it before I can jump on it now.

dnaq:
Hi, I'm on vacation at the moment so unfortunately I don't have access to
my developer environment. Will try to merge outstanding PRs next week.

JohnHeitmann:
Totally no rush, I was just making sure I wasn't holding things up. This isn't blocking me. Enjoy your vacation!

dnaq:
Hi,

I've been reviewing your code more thoroughly and my main issue right now (except for some whitespace and unnecessary imports that I can fix on my end) is the lack of tests, all test vectors from libsodium should be added, both for `aes256gcm` and for `chacha20poly1305`. Right now there are only three test vectors for `aes256gcm` and none for `chacha20poly1305`.

Also it would be nice if you could refactor the common tests `encrypt_decrypt` and `encrypt_decrypt_tamper` into the macro instead of duplicating them. `encrypt_decrypt_tamper` should also test the case where the authenticated data is tampered as well.

JohnHeitmann:
I beefed up aes256gcm with all vectors from libsodium, and improved the shared tests.

I don't see good tests in libsodium for chacha20poly1305. I found [this](https://github.com/jedisct1/libsodium/blob/22ab28be0a80878452836c08161a26005c460630/test/default/aead_chacha20poly1305.c), which has less coverage than the default aead tests in this PR, and I think has a few bugs (e.g. some failure branches fail on null when they should fail on truncation). If there's a place to grab better canned vectors point me at it.

JohnHeitmann:
Huh. Not sure why the build is failing. Works for me on both stable and nightly.

```
src/crypto/aead/aead_macros.rs:9:5: 9:20 error: unresolved import `rustc_serialize`. There is no `rustc_serialize` in `???` [E0432]
```

dnaq:
I've been looking at merging your pull request (I know it's been long overdue), however when looking at libsodium it seems that it nowadays implements two variants of chacha20poly1305 (the one from Adam Langleys RFC and another one). I'm not really a fan of having multiple mutually incompatible implementations of what is essentially the same primitive in the library, so if I would choose one of them I would choose the IETF one, on the other hand that might make things harder for people wanting compatibility with libsodium. What's your opinion on this issue?

I think we might do one of the following:

1. Remove chacha20poly1305 for now and only merge aes256gcm while we decide on the best way forward
2. Add chacha20poly1305_ietf and remove chacha20poly1305
3. Merge the pull request as is and add chacha20poly1305_ietf later

What do you think?

jedisct1:
I'd vote for 2)

dnaq:
@JohnHeitmann would you be interested in amending the pull request, removing support for chacha20poly1305 and adding support for chacha20poly1305_ietf (with the corresponding test vectors)?

briansmith:
If you remove the non-IETF variant then people can't implement the SSH protocol. If you remove the IETF variant than people can't implement TLS and most other protocols that use it.

dnaq:
@briansmith to be honest I don't see implementing TLS or implementing SSH as valid use cases for sodiumoxide. I'll have to think it over though, but NaCl (and thus libsodium and sodiumoxide by transitivity) had as its design goal that it should be easy to use, hard to misuse and that it should contain primitives that are good and secure for working programmers. A non-goal for the library was to be as flexible as possible, for that there's a multitude of other cryptographic libraries (and a multitude of examples where that flexibility has led to programmers implementing unsafe and dangerous crypto).

jedisct1:
Nod.

On the other hand, when people want something that is not available, they tend to write their own implementation based on what is available.

The situation is unfortunate, but both variants are used by common applications and protocols today.

briansmith:
> @briansmith to be honest I don't see implementing TLS or implementing SSH as valid use cases for sodiumoxide. I'll have to think it over though, but NaCl (and thus libsodium and sodiumoxide by transitivity)

I think that's a fair position. In that case, I recommend you pick the variant chosen by Noise, since Noise is basically trying to generalize NaCL. In particular, I think people are looking forward to implementing at least one of the Noise patterns using ChaCha20-Poly1305, using a NaCL-like API, so they don't need Salsa20.

dnaq:
That's a good point. I took a look at the noise paper, and it seems to use
the ietf version of the primitive.

JohnHeitmann:
I'll amend the request to use ..._ietf instead.

pijul:
I have implemented SSH (the crate is called thrussh), and I've called the C function directly, writing new bindings instead of using sodiumoxide, mostly to avoid extra vector allocations in the server side.

dnaq:
Hi. You never amended the pull request, so I'm going to close the issue if that's ok with you. Let me know if you want me to reopen it.

Issue #131, opened 2016-06-13
Use a specific version of libsodium


dirvine:
If possible I think that a very valid issue we could tackle is to use a specific version of libsodium and build it with a build.rs. This ensures we know what is built and tested with various tags etc. 

A secondary but very beneficial feature of this would be cross compiled of rust code is now very simple, but libsodium requires separate build script etc. If we combine this then use of sodiumoxide in cross compile becomes much simpler. 

If we did a PR for this, how would it be received?

dnaq:
I'm uncertain to be honest. When it comes to dynamic C-libraries I believe that they should be installed on the system and linked to (just like we are doing now). Otherwise if the underlying library has security issues we need to do a whole new update of `sodiumoxide` to not have the same issue. On the other hand I agree that the ergonomics could be better, somehow it would be nice if cargo allowed us to define external dependencies and print out a nicer error message if it's not found.

dirvine:
Yes I am thinking more static link here. For dynamic then not so sure. I am not all that comfortable with dynamic link and security software. I fear injection etc. I think if the underlying lib has a security issue then update is good actually as it helps take users to a new patched version of both (perhaps).  

I wonder if there can be some comprimise, like conditional compilation or similar, although that screws with crates.io really?

dnaq:
@dirvine If you find out a good way to solve this then please let me know and we'll discuss it further. Closing the issue for now.

dirvine:
Wll do, we are working on some trial implementations and looking first at a sodiumoxide_extras crate.  The intention being to check these things out (also seeded rng etc.) and from there look to get these PR's to you if you see them as useful. 

Thanks again for the work you do here. We will try and help where we can.

PR #135, opened 2016-07-11
Added feature to allow building or downloading libsodium.

Fraser999:
This adds a feature `get-libsodium` which will either download the prebuilt binaries if running MinGW on Windows, or clone libsodium and build from source if not on Windows (see #131).

Currently MSVC builds aren't handled for this feature, but if there is an interest in using this PR, we can probably add support for MSVC builds too if desired.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,1 +1,0 @@
-paths = ["libsodium-sys"]
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,4 @@
 # Cargo files
 Cargo.lock
 /target
-libsodium-sys/libsodium*
 libsodium-sys/target
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@
 # Cargo files
 Cargo.lock
 /target
+libsodium-sys/libsodium*
 libsodium-sys/target
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "libsodium-sys/libsodium"]
-	path = libsodium-sys/libsodium
-	url = https://github.com/jedisct1/libsodium.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libsodium-sys/libsodium"]
+	path = libsodium-sys/libsodium
+	url = https://github.com/jedisct1/libsodium.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ script:
 - cargo build --verbose
 - cargo test --verbose
 - cargo test --verbose --no-default-features
+- cargo test --verbose --features get-libsodium
 - cargo doc
 after_script:
 - cd target/doc
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ version = "0.0.12"
 
 [dependencies]
 libc = "0.2"
-libsodium-sys = "0.0.12"
-serde = { version="0.7", optional = true }
+libsodium-sys = { path = "libsodium-sys", version = "0.0.12" }
+serde = { version = "0.7", optional = true }
 
 [dev-dependencies]
 serde = "0.7"
@@ -22,3 +22,4 @@ rustc-serialize = "0.3"
 [features]
 benchmarks = []
 default = ["serde"]
+get-libsodium = ["libsodium-sys/get-libsodium"]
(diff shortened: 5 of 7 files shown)


jedisct1:
Very nice.

libsodium-sys/libsodium
```diff
@@ -0,0 +1 @@
+Subproject commit f01d7c128214e8f976889c8093a42d9cae1abe4b
```
mcginty: would it make more sense to "lock" the commit sodiumoxide chooses to link against to tagged releases? https://github.com/jedisct1/libsodium/releases (https://github.com/jedisct1/libsodium/commit/fce6852d64339efa33c0ee4130b3107b888d6067 is the latest, 1.0.10)

Fraser999: Yes, I was in two minds about this.

It would be good to have Windows and non-Windows builds using the exact same build of libsodium, and using the tagged commit would achieve that.

However, I was really keen to use the threadsafe version of `sodium_init()` which will be included in 1.0.11 since this still isn't handled by sodiumoxide, so I selfishly just used tip-of-tree of the stable branch of libsodium.

Ultimately I guess this will be a decision for @dnaq since he'll be the one updating the submodule commit and the tag version for Windows builds on a regular basis if this PR is accepted.  If the general preference is to use the tagged commit, we could probably reduce the maintenance overhead by defining the tag version outside of the platform-specific functions, then for non-Windows builds, do something like a `git fetch` then `git checkout $VERSION`.  That way, only the `VERSION` variable would need to be updated as and when a new release of libsodium lands.

mcginty: Ahh, I didn't know about what was coming in 1.0.11. Nifty!

I'd say from the conservative perspective, using specific stable releases whenever possible in cryptography libraries makes me do less of a double-take though (and it looks like the author signs releases, but not individual commits). I suppose this will be a non-issue once 1.0.11 is released officially.

dnaq: I'll just hold off on the merge until 1.0.11 is released. Thanks for all
your work.

mcginty: Looks like 1.0.11 is out now

Fraser999:
I've updated this a bit in the following ways:

1. Both Windows and non-Windows builds use the [libsodium release tarballs](https://download.libsodium.org/libsodium/releases/).  This avoids maintaining a git submodule - updating the libsodium version should be a single line change to the build script for all build types.  I did look at using the tagged commit as @mcginty suggested above, but the tagged commit hasn't had autoconf run on it - i.e. it's not straightforward to build from this commit.  It looks like the 'stable' branch is really the one intended for users building from source.
2. I changed how powershell downloads the tarball - the `wget` command I was using doesn't exist in the default powershell version on Windows 7.
3. On Windows, the doc tests fail to build if pthread is not explicitly linked (I'm not sure why building the doc tests is different to building the lib tests).  I added a section to the build script which guesses at the location of the MinGW pthread lib on the basis of the path to gcc.  This feels a bit flaky to me - it should maybe be a best-effort attempt, allowing sodiumoxide to still be built even though the doc tests would fail.

kpcyrd:
You've put a lot of effort into this, and I would suggest to have a separate package that supports something like this for simplicity.

I have some concerns with this, as it downloads binaries from an external server without signature verification (besides https, which doesn't help if the server is compromised or if somebody is able to generate valid certificates). Also, even if signature verification is added, some users might run into this accidentally and would prefer to have it fail and then install the necessary library from the repository of their distribution, instead of pulling executable binaries from the internet.

mcginty:
@kpcyrd It's not downloading binaries, it's downloading source and building (maybe you're thinking of the homebrew model ;)

As far as validating the code, how can this not be solved while keeping it in-package by simply "pinning" the libsodium signing key in this repo and then downloading the corresponding .sig file and verifying it's valid and from the known key? It can be an optional step and just print a warning that it couldn't validate if gpg isn't installed.

Also, curl has a `--pinnedpubkey` flag to be a little more rigid about what it's willing to download.

dnaq:
I know it's pretty late, but I've been looking at merging this pull request. However, it would be much nicer if we didn't have to hardcode the verison number of libsodium, and also if we could somehow verify that we have gotten the right tar-file before unpacking and installing.

I know that you maintain your own fork of sodiumoxide nowadays, so I don't know how interested you would be in updating this. If you're not, let me know so I can close the issue.

ffflorian:
So there is no chance this will get merged any time soon, right? :slightly_frowning_face:

dnaq:
Sorry, no it won't.

PR #138, opened 2016-08-02
Make vector manipulation more efficient.

brinchj:
This should not change the functionality of this function. It just modifies it to use constructs that are easier for the current Rust compiler to optimize (before specialization lands; not sure how much that will help).

I ran the benchmarks twice on my old and slow laptop to see if there were any consistent improvement. The small diffs seem to be part of the variation, but there does seem to be a significant gain for at least the xsalsa20poly1305 routine. Feel free to give it shot as well :-)

https://gist.github.com/brinchj/19b24ccdf34f07b7aaca22edf5e8dbfd

Note that drain() just does a memmove, so the removal of prefix-bytes is now inplace.

### DIFF
--- a/src/marshal.rs
+++ b/src/marshal.rs
@@ -8,12 +8,11 @@ pub fn marshal<T, F>(buf: &[u8],
                      ) -> (Vec<u8>, T)
     where F: Fn(*mut u8, *const u8, c_ulonglong) -> T {
     let mut dst = Vec::with_capacity(buf.len() + padbefore);
-    for _ in 0..padbefore {
-        dst.push(0);
-    }
-    dst.extend(buf.into_iter());
+    dst.resize(padbefore, 0u8);
+    dst.extend_from_slice(&buf[..]);
     let pdst = dst.as_mut_ptr();
     let psrc = dst.as_ptr();
     let res = f(pdst, psrc, dst.len() as c_ulonglong);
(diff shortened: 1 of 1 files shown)


dnaq:
Thanks. This is exactly what I wanted to do when first writing sodiumoxide, but back then Vec didn't have the functions for it. I will merge this after testing.

brinchj:
Sounds great! :+1:

jedisct1:
Sodiumoxide should really use `crypto_box_easy` and `crypro_secretbox_easy` API in order to completely avoid the padding dance.

dnaq:
Agreed.

brinchj:
If you think its not too complicated/risky, I can give migrating to *_easy APIs a shot. I agree that those APIs are what we want to use in the end.

brinchj:
Are there only cipher-anonymous easy APIs?

I do not see the variants that explicit call out the cipher configuration, so I am wondering whether these APIs are acceptable to use for cases that require backwards-compatibility (e.g. encrypted files). Can the underlying cipher change without the API changing? As I understand the NaCL APIs, the default choice can change in future versions (although I do not see any mention anywhere on backwards compatibility; which is why I assume it is out of scope).

jedisct1:
The underlying construction is not going to change without the API changing, or without a change to how the library is initialized.

brinchj:
That's good to know, thanks! :+1:

PR #144, opened 2016-11-23
Add randombytes_uniform function

dbrgn:
I added access to the [randombytes_uniform](https://download.libsodium.org/libsodium/content/generating_random_data/#usage) function.

Please make sure that everything is correct, this is my first FFI related Rust code :)

I'm not sure if there's a way to test the randomness of the data. Even asserting that the number is not the same as the previous 2 numbers could fail [in rare but valid cases](http://dilbert.com/strip/2001-10-25). Right now I just test 100 times that the number is smaller than the upper bound.

What we could maybe do is do it 1000 times and get the average. It should probably lie somewhere around `upper_bound / 2`.

### DIFF
--- a/libsodium-sys/src/randombytes.rs
+++ b/libsodium-sys/src/randombytes.rs
@@ -3,4 +3,5 @@
 extern {
     pub fn randombytes_buf(buf: *mut u8,
                            size: size_t);
+    pub fn randombytes_uniform(upper_bound: u32) -> u32;
 }
--- a/src/randombytes.rs
+++ b/src/randombytes.rs
@@ -26,3 +26,38 @@ pub fn randombytes_into(buf: &mut [u8]) {
         ffi::randombytes_buf(buf.as_mut_ptr(), buf.len());
     }
 }
+
+/// `randombytes_uniform()` returns an unpredictable value between 0 and
+/// `upper_bound` (excluded). It does its best to guarantee a uniform
+/// distribution of the possible output values.
+///
+/// THREAD SAFETY: `randombytes_uniform()` is thread-safe provided that you
+/// have called `sodiumoxide::init()` once before using any other function
+/// from sodiumoxide.
+pub fn randombytes_uniform(upper_bound: u32) -> u32 {
+    unsafe {
+        ffi::randombytes_uniform(upper_bound)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_randombytes_uniform_lt1000() {
+        for _ in 0..100 {
+            let random = randombytes_uniform(1000);
+            assert!(random < 1000);
+        }
+    }
+
+    #[test]
+    fn test_randombytes_uniform_lt3() {
+        for _ in 0..100 {
+            let random = randombytes_uniform(3);
+            assert!(random < 3);
+        }
+    }
+
+}
--- a/src/randombytes.rs
+++ b/src/randombytes.rs
@@ -53,6 +53,14 @@ mod test {
     }
 
     #[test]
+    /// Make sure that not all random numbers are the same.
+    fn test_randombytes_uniform_not_stuck() {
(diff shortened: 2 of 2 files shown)


dnaq:
Thanks for the pull request. The code seems to be fine and I don't think we need to sweat the tests that much (since I've read the libsodium source code). The tests could be remade a bit though, instead of testing that generated bytes are under a fixed value, do the testing in a loop and check to see that the generated value is less than the loop index. You might also want to add a test to see that the random generator isn't stuck, e.g. generate a bunch of values and test to see that all values aren't the same (might happen if linking messes up, not in any ordinary use-case).

The more important question is what the use case for this function is. Do users of sodiumoxide need to generate bounded u32s on any regular basis? The whole function seems a bit off, I would understand the need for a function that returned bounded bigints (even though that isn't really needed for sodiumoxide users it could be used for other cryptographic work), but when do we need to generate bounded u32s?

dbrgn:
> The tests could be remade a bit though, instead of testing that generated bytes are under a fixed value, do the testing in a loop and check to see that the generated value is less than the loop index.

Great idea.

> You might also want to add a test to see that the random generator isn't stuck, e.g. generate a bunch of values and test to see that all values aren't the same

Will add.

> The more important question is what the use case for this function is. Do users of sodiumoxide need to generate bounded u32s on any regular basis?

In my case it's to implement a form of PKCS#7 padding, where I need a random number in the (0, 255] range.

To quote the libsodium docs:

> The randombytes_uniform() function returns an unpredictable value between 0 and upper_bound (excluded). Unlike randombytes_random() % upper_bound, it does its best to guarantee a uniform distribution of the possible output values.

So apparently `randombytes_random` with a modulo doesn't seem to guarantee uniform distribution. I could pull in the `rand` crate to get a random number in that range, but why do that if libsodium already contains the necessary functionality?

dbrgn:
Tests are updated.

dnaq:
Are you aware that 'randombytes(1)' will return a vector with an u8 that has a uniformly distributed value between 0 and 255? Using modulo won't return a uniformly random value when the modulus isn't a power of two, e.g. when generating an RSA key.

dnaq:
By the way, pkcs#7 padding doesn't use random numbers.

dbrgn:
> Using modulo won't return a uniformly random value when the modulus isn't a power of two, e.g. when generating an RSA key.

Ah, that's what the docs are talking about! Maybe a clarification would help.

I still need a number between 1 (inclusive) and 255 (inclusive) though. Simply setting the LSB to 1 would again make the distribution non-uniform.

> By the way, pkcs#7 padding doesn't use random numbers.

Yes, I'm aware. It's not real PKCS#7, I'm implementing a [Threema Gateway](https://gateway.threema.ch/en) SDK where a random amount of PKCS#7 style padding (between 1 and 255 bytes with the value being the number of padding bytes) is added to an outgoing message.

dnaq:
Setting the LSB to one would eliminate all even numbers from your result.
The distribution would still be uniform though, but it wouldn't be the
distribution you're after. What you can do is to just draw random numbers,
see if the result is zero and if so redraw until it isn't. That's basically
what randombytes_uniform does.

dbrgn:
> Setting the LSB to one would eliminate all even numbers from your result.

True.

> The distribution would still be uniform though, but it wouldn't be the distribution you're after. What you can do is to just draw random numbers, see if the result is zero and if so redraw until it isn't.

That's right. But is there any reason against exposing this function if libsodium provides it?

dnaq:
I've just tried to keep the API surface as slim as possible to reduce the
learning curve of the library. Thinking of splitting it up in two libraries
in the future, one that only contains the base primitives of NaCl (+
ed25519 and maybe a password hashing primitive) and one crate that contains
the rest of libsodium.

I'll mull it over for a while, won't say that I won't add
randombytes_uniform, but it feels like a function that's only usable for
very rare cases.

PR #149, opened 2017-01-12
Support for the libsodium aeads (with ietf)

orthecreedence:
This is a continuation of the PR from @JohnHeitmann (https://github.com/dnaq/sodiumoxide/pull/130). I updated his code to have the `_ietf` versions of the chacha20poly1305 primitives. Tests passing locally.


dnaq:
Thanks. I will try to review and merge pretty soon.

libsodium-sys/src/crypto_aead_aes256gcm.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn crypto_aead_aes256gcm_npubbytes() -> size_t;
+    pub fn crypto_aead_aes256gcm_abytes() -> size_t;
+    pub fn crypto_aead_aes256gcm_encrypt(
+        c: *mut u8,
+        clen: *mut c_ulonglong,
+        m: *const u8,
+        mlen: c_ulonglong,
+        ad: *const u8,
+        adlen: c_ulonglong,
+        nsec: *mut [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: nsec is const unsigned char * in sodium

libsodium-sys/src/crypto_aead_aes256gcm.rs
```diff
@@ -0,0 +1,77 @@
...
+        k: *const [u8; crypto_aead_aes256gcm_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_aes256gcm_encrypt_detached(
+        c: *mut u8,
+        mac: *mut u8,
+        maclen: *mut c_ulonglong,
+        m: *const u8,
+        mlen: c_ulonglong,
+        ad: *const u8,
+        adlen: c_ulonglong,
+        nsec: *mut [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: nsec is const in sodium

libsodium-sys/src/crypto_aead_chacha20poly1305.rs
```diff
@@ -1,56 +1,77 @@
...
         clen: *mut c_ulonglong,
         m: *const u8,
         mlen: c_ulonglong,
         ad: *const u8,
         adlen: c_ulonglong,
-        nsec: *const [u8; crypto_aead_chacha20poly1305_NSECBYTES],
-        npub: *const [u8; crypto_aead_chacha20poly1305_NPUBBYTES],
-        k: *const [u8; crypto_aead_chacha20poly1305_KEYBYTES]) -> c_int;
-    pub fn crypto_aead_chacha20poly1305_decrypt(
+        nsec: *mut [u8; crypto_aead_chacha20poly1305_ietf_NSECBYTES],
```
dnaq: nsec is const in sodium

libsodium-sys/src/crypto_aead_chacha20poly1305.rs
```diff
@@ -1,56 +1,77 @@
...
+        k: *const [u8; crypto_aead_chacha20poly1305_ietf_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_chacha20poly1305_ietf_encrypt_detached(
+        c: *mut u8,
+        mac: *mut u8,
+        maclen: *mut c_ulonglong,
+        m: *const u8,
+        mlen: c_ulonglong,
+        ad: *const u8,
+        adlen: c_ulonglong,
+        nsec: *mut [u8; crypto_aead_aes256gcm_NSECBYTES],
```
dnaq: nsec is const in sodium

libsodium-sys/src/crypto_aead_chacha20poly1305.rs
```diff
@@ -1,56 +1,77 @@
...
+        npub: *const [u8; crypto_aead_aes256gcm_NPUBBYTES],
+        k: *const [u8; crypto_aead_aes256gcm_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_chacha20poly1305_ietf_decrypt_detached(
+        m: *mut u8,
+        nsec: *mut [u8; crypto_aead_aes256gcm_NSECBYTES],
+        c: *const u8,
+        clen: c_ulonglong,
+        mac: [u8; crypto_aead_aes256gcm_ABYTES],
```
dnaq: mac should be a pointer, right?

orthecreedence:
For wrapping C (I'm somewhat new to Rust's FFI), is the convention to use `*mut` if passing a pointer to an object that will be "filled in" by the called function? Or is it ok to use `*const` in this case if the API being wrapped uses it? For instance, in the last change requested

```rust
mac: [u8; crypto_aead_aes256gcm_ABYTES],
```

would be one of:

```rust
mac: *const [u8; crypto_aead_aes256gcm_ABYTES],
mac: *mut [u8; crypto_aead_aes256gcm_ABYTES],
```

Would `*const` be correct because the actual C API has `const` (`const unsigned char *mac`), or would it be `*mut` because the `mac` is filled in by the function?

dnaq:
In `libsodium-sys` we should bind to the exact API that libsodium provides, and your bindings differ at the moment. I will start to review your code again once the API matches exactly.

`nsec` is unused in libsodium, so that's why it's const.

When it comes to your question on the difference between `*const` and `*mut` in rust, I believe both are unsafe pointers, i.e. the compiler can't assume that something behind a *const won't mutate in unsafe code. The main difference is variance (see the [chapter on subtyping and variance](https://doc.rust-lang.org/nomicon/subtyping.html) in the Rustonomicon. For example `Vec<T>` uses `Unique<T>` which contains a `*const T` even though a `Vec<T>` can mutate it's contents (see [the chapter on Vec layout](https://doc.rust-lang.org/nomicon/vec-layout.html) in the same book.

orthecreedence:
Ok, thanks. That makes sense.

orthecreedence:
Any news on this? Just following up to make sure I haven't missed anything.

dnaq:
It's still on the to do list. Sorry that it has taken so long to merge.

orthecreedence:
No problem, looks like I missed a few things anyway.

libsodium-sys/src/crypto_aead_aes256gcm.rs
```diff
@@ -0,0 +1,77 @@
...
+        c: *const u8,
+        clen: c_ulonglong,
+        ad: *const u8,
+        adlen: c_ulonglong,
+        npub: *const [u8; crypto_aead_aes256gcm_NPUBBYTES],
+        k: *const [u8; crypto_aead_aes256gcm_KEYBYTES]) -> c_int;
+    pub fn crypto_aead_aes256gcm_encrypt_detached(
+        c: *mut u8,
+        mac: *mut u8,
```
dnaq: I think you missed this one...

dnaq: Nevermind. Took a look at libsodium, and the API is weird.

vvanders:
+1 on this, we've currently forked libsodium-sys since we needed the ietf of aead primitives.

edef1c:
@dnaq Is there a possibility of *just* getting the IETF AEADs in libsodium-sys sooner rather than later?

orthecreedence:
I tried merging the master branch into the AEAD branch a week or so ago and it seems the project has left the AEAD branch behind. So while I will continue the use my own version locally, I think at some point we'll need a new PR based off the master branch that provides the IETF primitives over the latest changes. I don't even care about GCM at this point, I just want the chacha/poly. If I get some time in the next few weeks I'll try to do this.

PR #150, opened 2017-01-16
implement seal_detached_inplace and open_detached_inplace

oconnor663:
https://github.com/dnaq/sodiumoxide/issues/24

This is another take on non-allocating primitives for secretbox. If we
like it, we could do the same thing for box too. There were two
particular design questions in my head for what interface to expose for
this:

1. Combined mode vs detached mode. Detached mode is the lower-level
primitive of the two, and it's easy to implement combined mode as a
wrapper, but not the other way around. This is also how libsodium's
implementation is organized. So I went with detached mode.

2. In-place vs two-buffers. Libsodium supports both, by checking the
pointers to see if they're overlapping. Rust makes it hard to support
both in a single interface, because &mut references can never be
aliased. I think it's easier to wrap an in-place function to become a
two-buffers one than the other way around, so I went with in-place.

Detached-and-in-place has the extra advantage that it can never panic,
because there are no length checks. The length of the tag is statically
known, and any length of the message is valid.

### DIFF
--- a/libsodium-sys/src/crypto_secretbox.rs
+++ b/libsodium-sys/src/crypto_secretbox.rs
@@ -17,6 +17,20 @@ extern {
         clen: c_ulonglong,
         n: *const [u8; crypto_secretbox_NONCEBYTES],
         k: *const [u8; crypto_secretbox_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_detached(
+        c: *mut u8,
+        mac: *mut [u8; crypto_secretbox_MACBYTES],
+        m: *const u8,
+        mlen: c_ulonglong,
+        n: *const [u8; crypto_secretbox_NONCEBYTES],
+        k: *const [u8; crypto_secretbox_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_open_detached(
+        m: *mut u8,
+        c: *const u8,
+        mac: *const [u8; crypto_secretbox_MACBYTES],
+        clen: c_ulonglong,
+        n: *const [u8; crypto_secretbox_NONCEBYTES],
+        k: *const [u8; crypto_secretbox_KEYBYTES]) -> c_int;
     pub fn crypto_secretbox_keybytes() -> size_t;
     pub fn crypto_secretbox_noncebytes() -> size_t;
     pub fn crypto_secretbox_macbytes() -> size_t;
--- a/src/crypto/box_/curve25519xsalsa20poly1305.rs
+++ b/src/crypto/box_/curve25519xsalsa20poly1305.rs
@@ -38,6 +38,13 @@ new_type! {
 }
 
 new_type! {
+    /// Authentication `Tag` for the detached encryption mode
+    ///
+    /// In the combined mode, the tag occupies the first MACBYTES bytes of the ciphertext.
+    public Tag(MACBYTES);
+}
+
+new_type! {
     /// `Nonce` for asymmetric authenticated encryption
     nonce Nonce(NONCEBYTES);
 }
@@ -89,6 +96,29 @@ pub fn seal(m: &[u8],
     c
 }
 
+/// `seal_detached()` encrypts and authenticates a message `m` using the senders secret key `sk`,
+/// the receivers public key `pk` and a nonce `n`. `m` is encrypted in place, so after this
+/// function returns it will contain the ciphertext. The detached authentication tag is returned by
+/// value.
+pub fn seal_detached(m: &mut [u8],
+                     &Nonce(ref n): &Nonce,
+                     &PublicKey(ref pk): &PublicKey,
+                     &SecretKey(ref sk): &SecretKey) -> Tag {
+    let mut tag = [0; MACBYTES];
+    unsafe {
+        ffi::crypto_box_detached(
+            m.as_mut_ptr(),
+            &mut tag,
+            m.as_ptr(),
+            m.len() as u64,
+            n,
+            pk,
+            sk,
+        );
+    };
+    Tag(tag)
+}
+
 /// `open()` verifies and decrypts a ciphertext `c` using the receiver's secret key `sk`,
 /// the senders public key `pk`, and a nonce `n`. It returns a plaintext `Ok(m)`.
 /// If the ciphertext fails verification, `open()` returns `Err(())`.
@@ -112,6 +142,33 @@ pub fn open(c: &[u8],
     };
     if ret == 0 {
         Ok(m)
+    } else {
+        Err(())
+    }
+}
+
+/// `open_detached()` verifies and decrypts a ciphertext `c` using the receiver's secret key `sk`,
+/// the senders public key `pk`, and a nonce `n`. `c` is decrypted in place, so if this function is
+/// successful it will contain the plaintext. If the ciphertext fails verification,
+/// `open_detached()` returns `Err(())`, and the ciphertext is not modified.
+pub fn open_detached(c: &mut [u8],
+                     mac: &Tag,
+                     &Nonce(ref n): &Nonce,
+                     &PublicKey(ref pk): &PublicKey,
+                     &SecretKey(ref sk): &SecretKey) -> Result<(), ()> {
+    let ret = unsafe {
+        ffi::crypto_box_open_detached(
+            c.as_mut_ptr(),
+            c.as_ptr(),
+            &mac.0,
+            c.len() as u64,
+            n,
+            pk,
+            sk,
+        )
+    };
+    if ret == 0 {
+        Ok(())
     } else {
         Err(())
     }
@@ -158,6 +215,26 @@ pub fn seal_precomputed(m: &[u8],
     c
 }
 
+/// `seal_detached_precomputed()` encrypts and authenticates a message `m` using a precomputed key
+/// `k` and a nonce `n`. `m` is encrypted in place, so after this function returns it will contain
+/// the ciphertext. The detached authentication tag is returned by value.
+pub fn seal_detached_precomputed(m: &mut [u8],
+                                 &Nonce(ref n): &Nonce,
+                                 &PrecomputedKey(ref k): &PrecomputedKey) -> Tag {
+    let mut tag = [0; MACBYTES];
+    unsafe {
+        ffi::crypto_box_detached_afternm(
+            m.as_mut_ptr(),
+            &mut tag,
+            m.as_ptr(),
+            m.len() as u64,
+            n,
+            k,
+        );
+    };
+    Tag(tag)
+}
+
 /// `open_precomputed()` verifies and decrypts a ciphertext `c` using a precomputed
 /// key `k` and a nonce `n`. It returns a plaintext `Ok(m)`.
 /// If the ciphertext fails verification, `open_precomputed()` returns `Err(())`.
@@ -179,6 +256,31 @@ pub fn open_precomputed(c: &[u8],
     };
     if ret == 0 {
         Ok(m)
+    } else {
+        Err(())
+    }
+}
+
+/// `open_detached_precomputed()` verifies and decrypts a ciphertext `c` using a precomputed key
+/// `k` and a nonce `n`. `c` is decrypted in place, so if this function is successful it will
+/// contain the plaintext. If the ciphertext fails verification, `open_detached()` returns
+/// `Err(())`, and the ciphertext is not modified.
+pub fn open_detached_precomputed(c: &mut [u8],
+                                 mac: &Tag,
+                                 &Nonce(ref n): &Nonce,
+                                 &PrecomputedKey(ref k): &PrecomputedKey) -> Result<(), ()> {
+    let ret = unsafe {
+        ffi::crypto_box_open_detached_afternm(
+            c.as_mut_ptr(),
+            c.as_ptr(),
+            &mac.0,
+            c.len() as u64,
+            n,
+            k,
+        )
+    };
+    if ret == 0 {
+        Ok(())
     } else {
         Err(())
     }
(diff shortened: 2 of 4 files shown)


src/crypto/secretbox/xsalsa20poly1305.rs
```diff
@@ -71,6 +71,25 @@ pub fn seal(m: &[u8],
     c
 }
 
+/// `seal_detached_inplace()` encrypts and authenticates a message `m` using a secret key `k` and a
+/// nonce `n`.  `m` is encrypted in place, so after this function returns it will contain the
+/// ciphertext.  The detached authentication tag is written to `mac`.
+pub fn seal_detached_inplace(m: &mut[u8],
```
dnaq: I'm not really happy with the API here. I think an api of the form
```
pub fn seal_detached_inplace(m: &mut [u8], &Nonce(ref n): &Nonce, &Key(ref k): &Key) -> Tag;
```

Where `Tag` is a newtype. This would be much more in spirit with the rest of the apis in sodiumoxide.

That would of course need to be complemented by a
```
pub fn seal_detached(m: &[u8], &Nonce(ref n), &Key(ref k)) -> (Vec<u8>, Tag);
```

I would like to deprecate the old `seal/open` apis in the future and only expose the new ones (renamed to seal/open).

oconnor663: I thought about using `Tag` at first, but the problem I ran into is that I don't know a safe way to get an `&Tag` from an `&[u8; MACBYTES]`. That's not a big deal if the tag is just sitting on the stack by itself, but if the caller wants it to be the first 16-bytes of a combined-mode buffer, I don't know if we'll be able to do that. Instead the caller/wrapper would have to do a second copy from their `Tag` on the stack to their buffer, after calling this function. (Or the other way around, for decryption.)

We could do a `transmute`, but that seemed like serious unsafety overkill, and I'm not even sure it's correct if the compiler chooses to do interesting things with padding in the newtype.

Any other ideas? Maybe having the caller do a copy isn't so bad, since we know it's only 64 bytes?

dnaq:
We expose from_slice for all newtypes in libsodium, we would do the same for Tag. For now from_slice does a copy, but that could change in the future.

oconnor663:
Added a commit with an API more like you suggested. At least one open question: should `open_detached_inplace` (or whatever we want to call it) take a `Tag` or a `&Tag`. By value feels cleaner, because `Tag` is `Copy`, but by reference would be more consistent with `&Nonce` and `&Key`?

dnaq:
Tags should be by reference, so that the APIs are consistent. Also all primitives should define their own Tag-types, don't reuse the one from crypto_onetimeauth. The actual primitive used for authentication is an implementation detail and shouldn't leak through the API.

oconnor663:
What do you think about the name, should we just call them `seal_detached` and `open_detached` and let the "inplace" part be implied?

dnaq:
Yes. Those names are better.

libsodium-sys/src/crypto_secretbox_xsalsa20poly1305.rs
```diff
@@ -16,12 +16,26 @@ extern {
         mlen: c_ulonglong,
         n: *const [u8; crypto_secretbox_xsalsa20poly1305_NONCEBYTES],
         k: *const [u8; crypto_secretbox_xsalsa20poly1305_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_detached(
+        c: *mut u8,
+        mac: *mut u8,
```
dnaq: shouldn't this be `mac: *mut [u8; crypto_secretbox_xsalsa20poly1305_MACBYTES]`?

libsodium-sys/src/crypto_secretbox_xsalsa20poly1305.rs
```diff
@@ -16,12 +16,26 @@ extern {
...
     pub fn crypto_secretbox_xsalsa20poly1305_open(
         m: *mut u8,
         c: *const u8,
         clen: c_ulonglong,
         n: *const [u8; crypto_secretbox_xsalsa20poly1305_NONCEBYTES],
         k: *const [u8; crypto_secretbox_xsalsa20poly1305_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_open_detached(
+        m: *mut u8,
+        c: *const u8,
+        mac: *const u8,
```
dnaq: Same here

oconnor663:
Added crypto_box functions.

oconnor663:
Any further thoughts on the `box_` and `secretbox` APIs?

dnaq:
I've scheduled some time this week to look at the patches further and
hopefully merge.

libsodium-sys/src/crypto_secretbox_xsalsa20poly1305.rs
```diff
@@ -16,12 +16,26 @@ extern {
         mlen: c_ulonglong,
         n: *const [u8; crypto_secretbox_xsalsa20poly1305_NONCEBYTES],
         k: *const [u8; crypto_secretbox_xsalsa20poly1305_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_detached(
```
dnaq: What are these functions doing in this file? They belong to crypto_secretbox.rs

libsodium-sys/src/crypto_secretbox_xsalsa20poly1305.rs
```diff
@@ -16,12 +16,26 @@ extern {
...
+        mlen: c_ulonglong,
+        n: *const [u8; crypto_secretbox_xsalsa20poly1305_NONCEBYTES],
+        k: *const [u8; crypto_secretbox_xsalsa20poly1305_KEYBYTES]) -> c_int;
     pub fn crypto_secretbox_xsalsa20poly1305_open(
         m: *mut u8,
         c: *const u8,
         clen: c_ulonglong,
         n: *const [u8; crypto_secretbox_xsalsa20poly1305_NONCEBYTES],
         k: *const [u8; crypto_secretbox_xsalsa20poly1305_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_open_detached(
```
dnaq: Same as above

libsodium-sys/src/crypto_secretbox_xsalsa20poly1305.rs
oconnor663: I don't think that's a file that exists. Should I create it?

I would've preferred to stick with the pattern in this file and import `crypto_secretbox_detached_xsalsa20poly1305`, but libsodium doesn't define that symbol. I wasn't sure whether it was better to split up the `extern` declarations or to keep them in one place and suck up with the naming inconsistency.

oconnor663: Ah my mistake, that file didn't exist when this PR was opened, but it does exist now. I'll rebase and move my declarations into there. (Do you have a preference for history-is-immutable merge commits vs rebase-and-squash?)

dnaq:
Doesn't matter for me, whatever is easiest for you.

oconnor663:
The latest squashed commit moves the extern declarations into the new `crypto_secretbox.rs` file. Are there any other style issues you noticed? I don't think I have any whitespace changes in the PR, but I could easily have missed something.

oconnor663:
Any other issues you've noticed, @dnaq?

dnaq:
Haven't had time to look any more lately. I'll try to take some time this
week.

libsodium-sys/src/crypto_secretbox.rs
```diff
@@ -17,6 +17,20 @@ extern {
         clen: c_ulonglong,
         n: *const [u8; crypto_secretbox_NONCEBYTES],
         k: *const [u8; crypto_secretbox_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_detached(
+        c: *mut u8,
+        mac: *mut [u8; crypto_secretbox_xsalsa20poly1305_MACBYTES],
```
dnaq: The wrong constants are used here, `crypto_secretbox_MACBYTES` and so on should be used in the file `crypto_secretbox.rs`.

src/crypto/box_/curve25519xsalsa20poly1305.rs
```diff
@@ -89,6 +96,29 @@ pub fn seal(m: &[u8],
...
 }
 
+/// `seal_detached()` encrypts and authenticates a message `m` using the senders secret key `sk`,
+/// the receivers public key `pk` and a nonce `n`. `m` is encrypted in place, so after this
+/// function returns it will contain the ciphertext. The detached authentication tag is returned by
+/// value.
+pub fn seal_detached(m: &mut [u8],
+                     &Nonce(ref n): &Nonce,
+                     &PublicKey(ref pk): &PublicKey,
+                     &SecretKey(ref sk): &SecretKey) -> Tag {
```
dnaq: I'm wondering if a better return type wouldn't be (&[u8], Tag), to improve the ergonomics of use. What do you think?

src/crypto/box_/curve25519xsalsa20poly1305.rs
```diff
@@ -117,6 +147,33 @@ pub fn open(c: &[u8],
...
 
+/// `open_detached()` verifies and decrypts a ciphertext `c` using the receiver's secret key `sk`,
+/// the senders public key `pk`, and a nonce `n`. `c` is decrypted in place, so if this function is
+/// successful it will contain the plaintext. If the ciphertext fails verification,
+/// `open_detached()` returns `Err(())`, and the ciphertext is not modified.
+pub fn open_detached(c: &mut [u8],
+                     mac: &Tag,
+                     &Nonce(ref n): &Nonce,
+                     &PublicKey(ref pk): &PublicKey,
+                     &SecretKey(ref sk): &SecretKey) -> Result<(), ()> {
```
dnaq: The change proposed above would also improve the ergonomics of the return value here. So that the return value would be `Result<&[u8], ()>`. (by the way I really need to get to adding real error types for sodiumoxide).

src/crypto/box_/curve25519xsalsa20poly1305.rs
```diff
@@ -158,6 +215,26 @@ pub fn seal_precomputed(m: &[u8],
     c
 }
 
+/// `seal_detached_precomputed()` encrypts and authenticates a message `m` using a precomputed key
+/// `k` and a nonce `n`. `m` is encrypted in place, so after this function returns it will contain
+/// the ciphertext. The detached authentication tag is returned by value.
+pub fn seal_detached_precomputed(m: &mut [u8],
+                                 &Nonce(ref n): &Nonce,
+                                 &PrecomputedKey(ref k): &PrecomputedKey) -> Tag {
```
dnaq: Same comment about the return type as above.

src/crypto/box_/curve25519xsalsa20poly1305.rs
```diff
@@ -184,6 +261,31 @@ pub fn open_precomputed(c: &[u8],
...
 }
 
+/// `open_detached_precomputed()` verifies and decrypts a ciphertext `c` using a precomputed key
+/// `k` and a nonce `n`. `c` is decrypted in place, so if this function is successful it will
+/// contain the plaintext. If the ciphertext fails verification, `open_detached()` returns
+/// `Err(())`, and the ciphertext is not modified.
+pub fn open_detached_precomputed(c: &mut [u8],
+                                 mac: &Tag,
+                                 &Nonce(ref n): &Nonce,
+                                 &PrecomputedKey(ref k): &PrecomputedKey) -> Result<(), ()> {
```
dnaq: Same comment about the return type as above.

src/crypto/secretbox/xsalsa20poly1305.rs
```diff
@@ -97,6 +124,31 @@ pub fn open(c: &[u8],
...
 }
 
+/// `open_detached()` verifies and decrypts a ciphertext `c` and and authentication tag `tag`,
+/// using a secret key `k` and a nonce `n`. `c` is decrypted in place, so if this function is
+/// successful it will contain the plaintext. If the ciphertext fails verification,
+/// `open_detached()` returns `Err(())`, and the ciphertext is not modified.
+pub fn open_detached(c: &mut[u8],
+                     tag: &Tag,
+                     &Nonce(ref n): &Nonce,
+                     &Key(ref k): &Key) -> Result<(), ()> {
```
dnaq: Same comment about the return type as above.

oconnor663:
Apologies for asking for feedback and then being slow to roll it in. Will get back to this PR later this week.

oconnor663:
I fixed the constants in `crypto_secretbox.rs` and rebased. Thoughts about returning the changed slice:

- Taking an `&mut [u8]` and returning a `&[u8]` can run into the awkward problem where the shared borrow keeps the mutable borrow alive, and you can't re-borrow the original buffer if you stash the (otherwise shared) return value. If we went with returning something, I think I'd want to return `&mut [u8]`.
- I'm not sure any of the existing callers would end up using this value. The likeliest place is probably here:

    ```rust
    let tag = seal_detached(&mut buf, &n, &k);
    open_detached(&mut buf, &tag, &n, &k).unwrap();
    ````

    Which would become either this (assuming we return `&mut[u8]`):

    ```rust
    let (c, tag) = seal_detached(&mut buf, &n, &k);
    open_detached(c, &tag, &n, &k).unwrap();
    ````

    Or this:

    ```rust
    let (_, tag) = seal_detached(&mut buf, &n, &k);
    open_detached(&mut buf, &tag, &n, &k).unwrap();
    ````

    I think the first (current) example is my favorite of the three. It's lighter on the punctuation, and it makes it more obvious that the exact same buffer is getting used both times. A returned slice might make more sense in an attached decryption function, where you're returning a sub-slice of the original buffer, but maybe not in these functions where you're always returning what the caller gave you?

oconnor663:
Any thoughts about those two bullet points?

dnaq:
I've thought it over, and I agree with your assessment. Will merge as soon
as I can find the time.

oconnor663:
Yay thanks!

PR #151, opened 2017-01-19
Add the missing “easy” functions for box/secretbox to libsodium-sys, and use them to implement open/seal in sodiumoxide

sopium:
“Completely avoid the padding dance”. Cc #138.


libsodium-sys/src/crypto_secretbox_xsalsa20poly1305.rs
```diff
@@ -22,6 +22,18 @@ extern {
         clen: c_ulonglong,
         n: *const [u8; crypto_secretbox_xsalsa20poly1305_NONCEBYTES],
         k: *const [u8; crypto_secretbox_xsalsa20poly1305_KEYBYTES]) -> c_int;
+    pub fn crypto_secretbox_easy(
```
dnaq: Shouldn't these additions be in crypto_secretbox.rs?

sopium:
I have run cargo bench and these changes made the secretbox seal/open bench slightly faster, by about 1.8%. The box seal/open bench seems to be dominated by pubkey operations though, and no obvious speedup can be observed.

libsodium-sys/src/crypto_secretbox_xsalsa20poly1305.rs
sopium: OK. I will add a `crypto_secretbox.rs` and move these two functions there.

sopium:
Well I run the benchmarks again and there is virtually no difference between the old and new code. I guess the vector manipulations were indeed well optimized. Kudos to @brinchj!

But at least the “easy” functions are not slower. The simplification is probably worth it.

PR #157, opened 2017-02-28
Remove .cargo/config, specify libsodium-sys path in Cargo.toml

### LINKED ISSUE #155: Trouble getting corresponding libsodium-sys using sodiumoxide from git(hub)
If `sodiumoxide` is specified in `Cargo.toml` as a dependency and points to some git repository, `libsodium-sys` is still retrieved from crates.io index.

Shall we get rid of `.cargo/config` and specify in `Cargo.toml` path of `libsodium-sys`?

I can send a PR if you think this is a good idea.

sopium:
This makes using sodiumoxide from git easier.

Fixes #155.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,1 +1,0 @@
-paths = ["libsodium-sys"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.0.14"
 
 [dependencies]
 libc = "0.2"
-libsodium-sys = "0.0.14"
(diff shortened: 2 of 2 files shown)


dnaq:
Does this work even if you add sodimoxide as a dependency from crates.io?

sopium:
Of course. See [cargo doc](http://doc.crates.io/specifying-dependencies.html#specifying-path-dependencies) here.

There are many projects out there specifying deps this way, e.g., openssl, gtk, (and my noise implementation 😴).

PR #160, opened 2017-04-30
Add key pair generation from seed also for box

chritchens:
It would be handy to be able to generate box key pairs by seed.Generating keys from seeds allows building key derivation functions also in the context of authenticated encryption.

The functionality is already in libsodium, and seed constants have already been added in libsodium-sys.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,12 @@ rust:
 - nightly
 sudo: false
 install:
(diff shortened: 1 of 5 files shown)


dnaq:
There are a lot of unrelated changes in this pull request (changes to .travis.yml, README and so on). I'm not opposed to the functionality, but I wont merge the pull request at the moment.

PR #161, opened 2017-04-30
Hide secret values in Debug output

pshc:
Hello! I noticed that `format!("{:?}", pk)` dumps the raw byte values even for `secret` newtypes. Seems safer to obscure those; what do you think?

### DIFF
--- a/src/newtype_macros.rs
+++ b/src/newtype_macros.rs
@@ -134,12 +134,6 @@ macro_rules! newtype_traits (($newtype:ident, $len:expr) => (
             b.index(_index)
         }
     }
-    impl ::std::fmt::Debug for $newtype  {
-        fn fmt(&self,
-               formatter: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-            write!(formatter, "{}({:?})", stringify!($newtype), &self[..])
-        }
-    }
     ));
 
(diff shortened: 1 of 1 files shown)


dnaq:
I see your point, but I'm a bit conflicted. I think `Debug` should only be used for debugging, and it's usable when designing a cryptosystem to be able to print out the intermediary values and keys to see if they are what you expect. On the other hand it should never be used in log files and so on, which is why the types don't implement `Display`. What do you think?

pshc:
I do agree! Ideally `Debug` should very rarely appear in production. But in practice, especially in larger teams, I think it will happen. I do sometimes use `Debug` in error handling, too.

I've done things like `println!(topLevelGodObject)` when getting familiar with a codebase, so I'm wary of this sort of situation, where a forgotten print statement generates a whole ton of output with sensitive data lodged in the middle.

So realistically, I think hiding the secret is a good default, considering the tradeoffs: Leaking a private key is really bad, whereas casting to a slice to see the raw numbers is very mildly inconvenient.

dnaq:
I think I'm ok with your reasoning. For the use-cases where intermediates
are needed the slice apis should be good enough.

pshc:
Thank you!!

PR #163, opened 2017-05-03
Support no_std

### LINKED ISSUE #159: no_std support
I'm interested in using sodiumoxide on microcontrollers. In my case I do have liballoc / libcollections available, so a fix to #24 is not necessary. However I'm wondering what additional work is required (investigating now).

kmcallister:
Fixes #159.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
 script:
 - cargo build --verbose
 - cargo test --verbose
-- if rustc --version | grep -q nightly; then cargo test --verbose --no-default-features; fi
+- if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then cargo test --verbose --no-default-features; fi
 - cargo test --verbose --no-default-features --features std
 - cargo doc
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 language: rust
+rust:
+- stable
+- beta
+- nightly
 sudo: false
 install:
 - wget https://github.com/jedisct1/libsodium/releases/download/1.0.8/libsodium-1.0.8.tar.gz
@@ -10,7 +14,8 @@ install:
 script:
 - cargo build --verbose
 - cargo test --verbose
-- cargo test --verbose --no-default-features
+- if rustc --version | grep -q nightly; then cargo test --verbose --no-default-features; fi
+- cargo test --verbose --no-default-features --features std
 - cargo doc
 after_script:
 - cd target/doc
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,5 +22,5 @@ rmp-serde = "0.12"
 
 [features]
 benchmarks = []
-no_std = []
-default = ["serde"]
+std = []
+default = ["serde", "std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,4 +22,5 @@ rmp-serde = "0.12"
 
 [features]
 benchmarks = []
+no_std = []
 default = ["serde"]
--- a/README.md
+++ b/README.md
@@ -36,6 +36,22 @@ differs between the C and Rust versions.
 Documentation for the latest build can be found at
 [gh-pages](https://dnaq.github.io/sodiumoxide).
 
+Optional features
+-----------------
+
+Several [optional features](http://doc.crates.io/manifest.html#usage-in-end-products) are available:
+
+* `std` (default: **enabled**). When this feature is disabled,
+  sodiumoxide builds using `#![no_std]`. Some functionality may be lost.
+  Requires a nightly build of Rust.
+
+* `serde` (default: **enabled**). Allows serialization and deserialization of
+  keys, authentication tags, etc. using the
+  [serde library](https://crates.io/crates/serde).
+
+* `benchmarks` (default: **disabled**). Compile benchmark tests. Requires a
+  nightly build of Rust.
+
 Examples
 --------
 TBD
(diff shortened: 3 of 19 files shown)


dnaq:
Thanks. Will review and merge as soon as I have time.

kmcallister:
@dnaq: poke?

kmcallister:
@dnaq: I need this for an important project. Are you actually interested in the patch or should I maintain my own fork of sodiumoxide?

dnaq:
Hopefully this weekend

dnaq:
Just tried to merge, but `#[feature]` isn't supported in stable, only in nightly. I'm a bit wary of adding features that only build in nightly. Do you know when these features will land in stable?

I'm not at all opposed to you vendoring your dependencies if need be, and I'm grateful for your patch, but I think compatibility with stable rust is quite important. The only exception so far has been to run benchmarks, which isn't in any way core functionality.

A solution might be to only build the non-allocating primitives when using `no_std`, then you wouldn't need `collections` and `alloc`. Would that work for you, and if so are you interested in amending your patch?

kmcallister:
I don't see the problem. Many crates have optional features that require nightly. Every piece of code in sodiumoxide today still builds on stable. CI will confirm this. Users who don't care about no_std should see no difference at all; if that's not the case, I consider it a bug and will endeavour to fix it.

I do have the concern that we're not testing the no_std option in CI. If you want me to fix that before this PR lands, I can try to do that.

Someone did tell me that it's better to have a Cargo feature "std" that *is* on by default. (Cargo features should never *remove* functionality, and that could concievably happen with no_std, although it's not the case in the initial patch.) So I'll make that change when I get back from the woods on Sunday.

> A solution might be to only build the non-allocating primitives when using no_std, then you wouldn't need collections and alloc. Would that work for you, and if so are you interested in amending your patch?

Doesn't help; no_std itself is unstable. Besides, it's a lot more work and probably not necessary for my use case.

> I'm not at all opposed to you vendoring your dependencies if need be

I am *very* hesitant to mantain my own patches on top of a cryptography library. We all know how well that went for Debian.

dnaq:
> Doesn't help; no_std itself is unstable.

I believe that it's possible to create `#![no_std]` libraries in rust, it's just binaries that are unstable at the moment. The [book](https://doc.rust-lang.org/book/no-stdlib.html) seems to agree with me. On the other hand anyone who wants to actually use sodiumoxide without std would have to be running nightly at the moment, so alloc and friends might not be a blocker.

> I do have the concern that we're not testing the no_std option in CI. If you want me to fix that before this PR lands, I can try to do that.

This is one of my concerns, and any work you could put towards making the tests run automatically would really help.

> I am very hesitant to mantain my own patches on top of a cryptography library. We all know how well that went for Debian.

That's a quite valid view. The reason that I have been holding off on a lot of merges lately is that I never merge anything into sodiumoxide without reviewing it and testing it. Lately I've been very short on time, but I've been taking the view that a stable and thoroughly vetted cryptographic library is more important for the rust ecosystem than a library with all latest bells and whistles. I know these might sound like big words for what is essentially a bindings library, but there are a lot of cryptographic libraries with weaknesses and weird behaviour, and I would like to avoid sodiumoxide becoming one.

That being said, embedded systems have been a target from the beginning, and `#![no_std]` is work towards that goal, so I would really like to merge your work when the CI-bit has been solved. Your patch is pretty small, and I can hopefully merge it by next weekend.

kmcallister:
Ah, I didn't know that no_std is stable for libraries! I didn't get an
"unused #[feature]" warning; that might be a bug in rustc.

I totally agree about being careful. A lot of danger can creep in at the
API level even if the underlying primitives are sound. In this case though,
there is no API change.

So I'll switch to #[cfg(not(feature = "std"))], and also work on CI
integration. Should be ready in a day or two. Thanks!

On May 21, 2017 11:40 AM, "dnaq" <notifications@github.com> wrote:

Doesn't help; no_std itself is unstable.

I believe that it's possible to create #![no_std] libraries in rust, it's
just binaries that are unstable at the moment. The book
<https://doc.rust-lang.org/book/no-stdlib.html> seems to agree with me. On
the other hand anyone who wants to actually use sodiumoxide without std
would have to be running nightly at the moment, so alloc and friends might
not be a blocker.

I do have the concern that we're not testing the no_std option in CI. If
you want me to fix that before this PR lands, I can try to do that.

This is one of my concerns, and any work you could put towards making the
tests run automatically would really help.

I am very hesitant to mantain my own patches on top of a cryptography
library. We all know how well that went for Debian.

That's a quite valid view. The reason that I have been holding off on a lot
of merges lately is that I never merge anything into sodiumoxide without
reviewing it and testing it. Lately I've been very short on time, but I've
been taking the view that a stable and thoroughly vetted cryptographic
library is more important for the rust ecosystem than a library with all
latest bells and whistles. I know these might sound like big words for what
is essentially a bindings library, but there are a lot of cryptographic
libraries with weaknesses and weird behaviour, and I would like to avoid
sodiumoxide becoming one.

That being said, embedded systems have been a target from the beginning,
and #![no_std] is work towards that goal, so I would really like to merge
your work when the CI-bit has been solved. Your patch is pretty small, and
I can hopefully merge it by next weekend.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/dnaq/sodiumoxide/pull/163#issuecomment-302954662>, or mute
the thread

kmcallister:
@dnaq: OK I think this is ready, let me know if you have any remaining comments.

kmcallister:
Sweet! Thanks. Now if only I hadn't bricked my Cortex-M dev board...

PR #167, opened 2017-07-06
Build libsodium from sources

oherrala:
This pull request modifies the `libsodium-sys`'s `build.rs` to do the following:

1. Try if pkgconfig can find libsodium and use that if found.
2. Get libsodium using Git (this PR uses libsodium tag 1.0.12), compile it and link `libsodium-sys` against it (static linking).

In the future, when new libsodium version is released, the git submodule needs to be updated to point into new version.

This PR also drops support for environment variables `SODIUM_LIB_DIR` and `SODIUM_STATIC`. However, these are easy to add back if so needed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libsodium-sys/libsodium"]
+	path = libsodium-sys/libsodium
+	url = https://github.com/jedisct1/libsodium
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,6 @@ rust:
 - beta
 - nightly
 sudo: false
-install:
-- wget https://github.com/jedisct1/libsodium/releases/download/1.0.8/libsodium-1.0.8.tar.gz
-- tar xvfz libsodium-1.0.8.tar.gz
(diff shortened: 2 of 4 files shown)


jedisct1:
Maybe use the stable branch instead.

oherrala:
@jedisct1 

> Maybe use the stable branch instead.

Using stable branch is also option. I picked the git tag to get more deterministic build so that changes in libsodium (master or some branch) doesn't directly affect sodiumoxide. This way users building the same crate version would get the same result (except when pkgconfig finds libsodium).

antiochp:
Hey - what's the status of this? Would be awesome to be able to use the `sodiumoxide` crate like this without depending on libsodium being installed on the system.
Specifically - trusty on travisci.

dnaq:
Thanks for the pull-request. The functionality is nice, but I'm a bit wary of downloading third party sources and building them as part of the install of a rust-package. I'm still undecided though.

dnaq:
I'm closing this for now. Hopefully there will be a nice way to integrate third party sources in the future.

PR #174, opened 2017-10-19
Create libsodium-sys automatically using bindgen

dfaust:
Creating libsodium-sys automatically eliminates the work of doing that manually, thereby allowing developers to spend more time on designing the high-level API of sodiumoxide. It also grants access to all new features of libsodium as they are released by using libsodium-sys directly.

The types used by `bindgen` differ slightly from those currently used in `libsodium-sys`, so a migration is necessary:

* `randombytes_buf` and `sodium_memzero` now take `*const ::std::os::raw::c_void` values. Pointers passed to those functions must be cast using `as *const ::std::os::raw::c_void`.

* Many functions now take `*const u8` / `*mut u8` for readable / writable buffers. So instead of passing `&buf` / `&mut buf` they need to be passed `buf.as_ptr()` / `buf.as_mut_ptr()`.

* Some functions like `crypto_pwhash_scryptsalsa208sha256` take `*const c_char` parameters, since `c_char` is defined as `i8`, the pointers can be cast with `as *const i8`.

* Static strings like `crypto_box_PRIMITIVE` are directly imported from `libsodium` which means they are nul-terminated. They can be read using `std::ffi::CStr::from_bytes_with_nul`. And of cause they can be exported in `sodiumoxide` as a `&str` (like it is done with `crypto_pwhash_scryptsalsa208sha256_STRPREFIX`).


dnaq:
Hi and thanks for the pull request. This is good work that I've been thinking of implementing for a while. Unfortunately a bug in bindgen blocks the merge for now. The bug is that bindgen erroneously infers constness of pointers. e.g. for `void sodium_memzero(void *const pnt, const size_t len)` bindgen generates the rust following `pub fn sodium_memzero(sp *const ::std::os::raw::c_void, len: usize);`.

See [bindgen issue 511](https://github.com/rust-lang-nursery/rust-bindgen/issues/511) for more information. I hesitate on merging until that bindgen issue is resolved.

Still, thanks for your work. It's much appreciated. Is it ok if you resubmit a new pull-request when the bindgen issue is resolved?

dfaust:
That is unfortunate, I will re-submit a PR when the bug is fixed.
Personally, I need some newer functions from libsodium and can live with a little bit of const casting.

PR #176, opened 2017-10-26
Add `minmal` feature analogue to libsodium's `--enable-minimal`

dfaust:
`sodiumoxide` needs to be built with the `minmal` feature enabled when linking against libsodium, built with the `--enable-minimal` flag.

I don't like defining a feature that removes functionality but in this case I wanted it to be analogous with the libsodium flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,4 +24,5 @@ rmp-serde = "0.13"
 [features]
 benchmarks = []
 std = []
-default = ["serde", "std"]
+minimal = []
+default = ["serde", "std", "minimal"]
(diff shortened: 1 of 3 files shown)


dnaq:
Thanks for the pull request. I might consider removing the deprecated functions from sodiumoxide as well instead of adding a new feature. What do you think?

dfaust:
I would say, at this stage there probably aren't many users of sodiumoxide using these deprecated functions. But going forward, this likelihood will only increase, as more functions get deprecated.

Removing these functions from sodiumoxide would be a disruption. Although I guess it would be acceptable to ask developers to use the libsodium-sys interface directly, so the functionality could be removed from sodiumoxide. But it should still be kept in libsodium-sys. This of cause would work automatically if libsodium-sys was generated with bindgen.

dnaq:
The plan is to generate libsodium-sys with bindgen as soon as bindgen has fixed some bugs. See #174 for more info.

dfaust:
I'm closing this PR since it turns out `crypto_stream_aes128ctr` was removed completely from libsodium. So #178  should be merged instead.

PR #178, opened 2017-10-31
Remove aes128ctr (Fix #177)

### LINKED ISSUE #177: aes128ctr was removed from libsodium since 1.0.15
aes128ctr was removed by [this commit](https://github.com/jedisct1/libsodium/commit/96be673f82ed7068046cb80584e257f5b81572a4#diff-90c54a2f5f88237dba59c6231015b58f).

It prevents sodiumoxide from building when libsodium 1.0.15 is installed:

```
  = note: Undefined symbols for architecture x86_64:
            "_crypto_stream_aes128ctr_xor", referenced from:
                sodiumoxide::crypto::stream::aes128ctr::stream_xor::h131f4dac97323cf8 in sodiumoxide-862d462817794df8.sodiumoxide8.rust-cgu.o
                sodiumoxide::crypto::stream::aes128ctr::stream_xor_inplace::hc3b89f169d4b760c in sodiumoxide-862d462817794df8.sodiumoxide8.rust-cgu.o
            "_crypto_stream_aes128ctr", referenced from:
                sodiumoxide::crypto::stream::aes128ctr::stream::hdf78b0a33104be31 in sodiumoxide-862d462817794df8.sodiumoxide8.rust-cgu.o
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)


error: aborting due to previous error

error: Could not compile `sodiumoxide`.
```

kcchu:
aes128ctr had been removed from libsodium in 1.0.15.


dfaust:
You can find an alternate solution at #176 .

kcchu:
Thanks. #176 does work around the build issue. However, the flag it proposed isn’t very intuitive. Since libsodium removed the code entirely, because the implementation was “slow, non-standard, not
authenticated”, it make sense for sodiumoxide to follow suit and remove it too.

dfaust:
Oh, you're right. In that case it really should be removed.

PR #179, opened 2017-11-22
Unbreak Build

naftulikay:
Do it for the internet points.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ rust:
 - stable
 - beta
 - nightly
+matrix:
+  allow_failures:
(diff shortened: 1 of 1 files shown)


dnaq:
In what way does your patch unbreak the build? I can run all tests on nightly as of now, is there any specific case where the nightly build breaks?

naftulikay:
Master has a red badge. Make sure to clear your cache.

dnaq:
I fixed the underlying issue instead. Thanks for bringing it to my attention.

naftulikay:
👍

PR #180, opened 2017-12-01
Adding the chacha20-poly1305 construction (standard version)

P-E-Meunier:
So, I'm thinking of using SodiumOxide in Thrussh, my SSH library, but I'd need this PR to be merged and on crates.io relatively soon, is this doable?

### DIFF
--- a/src/crypto/aead/chacha20poly1305.rs
+++ b/src/crypto/aead/chacha20poly1305.rs
@@ -1,37 +1,19 @@
-//! The original NaCl variant of the Chacha20-Poly1305 construction,
-//! an algorithm for Authenticated Encryption with Additional Data
+//! The original variant of the Chacha20-Poly1305 construction, an
+//! algorithm for Authenticated Encryption with Additional Data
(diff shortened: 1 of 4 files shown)


dnaq:
Thanks for the pull request. I might be able to merge this weekend, but I
need some test vectors and unit tests as well. Take a look at the tests for
the other cryptographic primitives for reference. For each test vector you
add please also add a source to where the test vector is from.

src/crypto/aead/chacha20poly1305.rs
```diff
@@ -0,0 +1,187 @@
+//! The original NaCl variant of the Chacha20-Poly1305 construction,
+//! an algorithm for Authenticated Encryption with Additional Data
+//! (AEAD).
+
```
dnaq: The documentation here is a bit misleading. NaCl doesn't contain Chacha20-Poly1305.

dnaq:
Implemented in 5d53fab116011dba7fd32eedf684b1e1106743d4

PR #208, opened 2018-05-12
Introduce and generate libsodium bindings using rust-bindgen

cynecx:
The transition to rust-bindgen actually went really smooth, besides some "issues" like the hardcoded `crypto_pwhash_scryptsalsa208sha256_STRPREFIX` constant, everything else seems trivial.

Most of the changes were:

- Use `as_ptr()` and `as_mut_ptr()` instead of relying on rust-ized fitting types on the bindings part
- bindgen generates constants with `u32` as type, hence all the type casts to `usize`
- While at it, rustfmt the tests in libsodium-sys
- Moved libsodium-sys tests into their own modules
- Introduced new env variable: `SODIUM_INC_DIR`

PR #210, opened 2018-05-14
Add binding for sodium_add function

kurnevsky:
It's useful for incrementing nonces with some step value.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -31,16 +31,37 @@ pub fn memcmp(x: &[u8], y: &[u8]) -> bool {
     }
 }
 
-/// `increment_le()` treats `x` as an unsigned little-endian number and increments it.
+/// `increment_le()` treats `x` as an unsigned little-endian number and increments it in
+/// constant time.
 ///
 /// WARNING: this method does not check for arithmetic overflow. When used for incrementing
-/// nonces it is the callers responsibility to ensure that any given nonce value
-/// is only used once.
+/// nonces it is the caller's responsibility to ensure that any given nonce value
+/// is used only once.
 /// If the caller does not do that the cryptographic primitives in sodiumoxide
-/// will not uphold any security guarantees (i.e. they will break)
+/// will not uphold any security guarantees (i.e. they may break)
 pub fn increment_le(x: &mut [u8]) {
     unsafe {
         ffi::sodium_increment(x.as_mut_ptr(), x.len());
+    }
+}
+
+/// `add_le()` treats `x` and `y` as unsigned little-endian numbers and adds `y` to `x`
+/// modulo 2^(8*len) in constant time.
+///
+/// `add_le()` will return Err<()> if the length of `x` is not equal to the length of `y`.
+///
+/// WARNING: When used for incrementing nonces it is the caller's responsibility to ensure
+/// that any given nonce value is used only once.
+/// If the caller does not do that the cryptographic primitives in sodiumoxide
+/// will not uphold any security guarantees (i.e. they may break)
+pub fn add_le(x: &mut [u8], y: &[u8]) -> Result<(), ()> {
+    if x.len() == y.len() {
+        unsafe {
+            ffi::sodium_add(x.as_mut_ptr(), y.as_ptr(), x.len());
+        }
+        Ok(())
+    } else  {
+        Err(())
     }
 }
 
@@ -52,7 +73,7 @@ mod test {
     fn test_memcmp() {
         use randombytes::randombytes;
 
-        for i in 0usize..256 {
+        for i in 0..256 {
             let x = randombytes(i);
             assert!(memcmp(&x, &x));
             let mut y = x.clone();
@@ -72,7 +93,7 @@ mod test {
 
     #[test]
     fn test_increment_le_zero() {
-        for i in 1usize..256 {
+        for i in 1..256 {
             let mut x = vec![0u8; i];
             increment_le(&mut x);
             assert!(!x.iter().all(|x| *x == 0));
@@ -113,10 +134,78 @@ mod test {
 
     #[test]
     fn test_increment_le_overflow() {
-        for i in 1usize..256 {
+        for i in 1..256 {
             let mut x = vec![255u8; i];
             increment_le(&mut x);
             assert!(x.iter().all(|xi| *xi == 0));
+        }
+    }
+
+    #[test]
+    fn test_add_le_zero() {
+        for i in 1..256 {
+            let mut x = vec!(0u8; i);
+            let mut y = vec!(0u8; i);
+            y[0] = 42;
+            assert!(add_le(&mut x, &y).is_ok());
+            assert!(!x.iter().all(|x| { *x == 0 }));
+            assert_eq!(x, y);
+        }
+    }
+
+    #[test]
+    fn test_add_le_vectors() {
+        let mut x = [255, 2, 3, 4, 5];
+        let y = [42, 0, 0, 0, 0];
+        let z = [41, 3, 3, 4, 5];
+        assert!(add_le(&mut x, &y).is_ok());
+        assert!(!x.iter().all(|x| { *x == 0 }));
+        assert_eq!(x, z);
+        let mut x = [255, 255, 3, 4, 5];
+        let z = [41, 0, 4, 4, 5];
+        assert!(add_le(&mut x, &y).is_ok());
+        assert!(!x.iter().all(|x| { *x == 0 }));
+        assert_eq!(x, z);
+        let mut x = [255, 255, 255, 4, 5];
+        let z = [41, 0, 0, 5, 5];
+        assert!(add_le(&mut x, &y).is_ok());
+        assert!(!x.iter().all(|x| { *x == 0 }));
+        assert_eq!(x, z);
+        let mut x = [255, 255, 255, 255, 5];
+        let z = [41, 0, 0, 0, 6];
+        assert!(add_le(&mut x, &y).is_ok());
+        assert!(!x.iter().all(|x| { *x == 0 }));
+        assert_eq!(x, z);
+        let mut x = [255, 255, 255, 255, 255];
+        let z = [41, 0, 0, 0, 0];
+        assert!(add_le(&mut x, &y).is_ok());
+        assert!(!x.iter().all(|x| { *x == 0 }));
+        assert_eq!(x, z);
+    }
+
+    #[test]
+    fn test_add_le_overflow() {
+        for i in 1..256 {
+            let mut x = vec!(255u8; i);
+            let mut y = vec!(0u8; i);
+            y[0] = 42;
+            assert!(add_le(&mut x, &y).is_ok());
+            assert!(!x.iter().all(|x| { *x == 0 }));
+            y[0] -= 1;
+            assert_eq!(x, y);
+        }
+    }
+
+    #[test]
+    fn test_add_le_different_lengths() {
+        for i in 1..256 {
+            let mut x = vec!(1u8; i);
+            let y = vec!(42u8; i + 1);
+            let z = vec!(42u8; i - 1);
+            assert!(add_le(&mut x, &y).is_err());
+            assert_eq!(x, vec!(1u8; i));
+            assert!(add_le(&mut x, &z).is_err());
+            assert_eq!(x, vec!(1u8; i));
         }
     }
 }


nvesely:
This should be rebased after #208 is merged. I'm in favor of keeping the original function name, `sodium_add`. While arguably a less descriptive name, one can more easily cross-reference the libsodium documentation when working with sodiumoxide.

kurnevsky:
@nvesely I chose the name to be consistent with `sodium_increment` - its binded to `increment_le`. And other bindings (`memzero`, `memcmp`) don't have prefix `sodium_`. So you still think we should use `sodium_add` name regardless of what we already have?

dnaq:
I'm ok with the name being `add_le()` for consistency. I would also like all `nonce` newtypes to have methods with the following signatures:
```
fn add_le(&self, other: &$name) -> $name;
fn add_le_inplace(&mut self, other: &$name)
```

kurnevsky:
@dnaq is there practical sense in adding one nonce to another? As I see the use case with `add_le` - we usually want to add some integer value that we calculated somehow. But adding another nonce will mean that we have to construct nonce from our integer value.

Tom-Phinney:
A nonce is a value that supposedly has not been used before in its particular sequence. For example, a sequence of alternate-sign integers, S = {`+1`, `-2`, `+3`, `-4`, `+5`, `-6`, … }, all `as u<N>`, can serve as a sequence of nonces for a cryptographic association.

The sum or difference of elements drawn from two nonce sequences do not necessarily form a third nonce sequence. For example, S+S would lead to multiple occurrences of `+1` (e.g., `(-4)+(+5)`, `(+5)+(-4)`, `(+7)+(-6)`, etc).

I see no reason to add a binding that encourages users to potentially violate the nonce postulate, which would be that the value produced by the addition has not previously occurred in the sequence.

kurnevsky:
@Tom-Phinney to clarify a bit - you are speaking about adding methods to `Nonce` types, right? Not about adding a binding for `sodium_add` function?

dnaq:
@Tom-Phinney @kurnevsky The idea was to add a binding that helps users that want to increment the nonce by a fixed amount, the idea wasn't that people should add arbitrary nonces together. That being said I agree that it might be bad idea, so we'll skip those methods.

kurnevsky:
@dnaq we can still add method to increment the nonce by 1. Regarding increment the nonce by a fixed amount - it's unclear which type we should use for this amount.

naftulikay:
I am also trying to find a legitimate use case for providing nonce addition. Can't think of one that isn't dangerous potentially.

Tom-Phinney:
I have no problem with incrementing a nonce by a constant. Many protocols use `u<N>` as a nonce sequence, so `nonce += 1u<N>` makes sense. Some comm protocols use the time of message formation, at some granularity and with respect to a reference point, as the nonce. The current time gets incremented, leading to a new nonce.

My only concern was that we not support the notion that adding arbitrary nonces from two sequences will produce a third nonce sequence. The example above shows that is not the case.

dnaq:
We might think about adding a method:

‘’’
fn add_le(&self, num: &[u8]) -> Nonce
‘’’

in the future, but that is still out of scope for this pull request.

Dylan-DPC:
@Tom-Phinney @dnaq thoughts about this? can we merge it?

Tom-Phinney:
It's okay by me to merge. Since Nonce is a `[u8]`, adding a `&[u8]` is general enough to permit any operand up to the storage size of the nonce. 

Has any thought been given to the RHS operand name? I often find these somewhat arbitrary, and thus beyond intuition. `other` is the operand name used in the normal `add` and `wrapping_add` operator definitions. Of course those are names for a second operand of the same type as the first, which is not the case here.

Alternatively, a more descriptive name such as `inc` might be used. My concern about the suggested `num` is that `num` implies a scalar to me, rather than a slice of byte-size `u8`s.

nvesely:
> @nvesely I chose the name to be consistent with sodium_increment - it's binded to increment_le. And other bindings (memzero, memcmp) don't have prefix sodium_. So you still think we should use sodium_add name regardless of what we already have?

I think we should aim to provide an interface as close to identical as sodiumoxide does. So, yes. And I think we should change many other names and the modular organization of the repo as well, but that's for another PR.

nvesely:
To address the discussion of `Nonce`s, I'm against the user ever really dealing with them. This is something that should be abstracted away as much as possible because that's safer and there's not really a good reason afaict a user would need to deal with them at all. In my opinion, generally speaking a cryptography library should not even include a public nonce type. Instead, I would just transparently generate the nonce during encryption and encode it in the ciphertext.

To give a little more background to what @Tom-Phinney wrote, a nonce is what provides CPA-security. The advantage of an adversary in breaking the CPA-security of a stream cipher stream in the random oracle model is directly related to the probability that any two queries the adversary makes are answered with nonce values that overlap (the standard way a nonce is chosen is to pick a value uniformly at random in the set of all byte strings in the length of the security parameter n, and then use that to encrypt n bytes of the plaintext). When n is sufficiently large, any PPT adversary will only observe an overlap in their queries with negligible probability.

Less theoretically, one problem would be that if the same message is encrypted twice using the same nonce, it will be possible to see that it's the same message (here and above btw "message" refers to `n` bytes and not the entire input given for encryption). If no nonce is used, this observation will be possible among all messages. Either way, this is *not good*™.

> Some comm protocols use the time of message formation, at some granularity and with respect to a reference point, as the nonce. The current time gets incremented, leading to a new nonce.

This appears to be an objectively incorrect thing to do. Can you provide some examples of protocols that do this in the stream cipher context? The proof that turns a PRF into a CPA-secure private-key encryption scheme relies on the nonce being chosen uniformly at random and a timestamp is not a uniformly-at-random chosen value.

One thing that's cool about stream ciphers is the ability to decrypt an arbitrary portion of a ciphertext. This is because they just generate a pseudorandom byte stream that is XORed with the plaintext, and no portion of this stream is dependent on any other. So you can simply generate just the pseudorandom bytes you need and use them to decrypt just the plaintext bytes you need to read. If you know that your ciphertext is a fixed-format message, it could allow a huge reduction in computation for someone using sodiumoxide if they were able to specify a byte range to decrypt. In this case they still need not deal with the nonce directly. They just pass the key, ciphertext, start, and end, and we perform the nonce addition for them. This is the only case where simply "inc"ing the nonce is not appropriate thing to do.

src/utils.rs
```diff
@@ -44,6 +44,24 @@ pub fn increment_le(x: &mut [u8]) {
...
+/// `add_le()` will panic if length of `x` is not equal to length of `y`.
+///
+/// WARNING: this method does not check for arithmetic overflow. When used for incrementing
+/// nonces it is the callers responsibility to ensure that any given nonce value
+/// is only used once.
+/// If the caller does not do that the cryptographic primitives in sodiumoxide
+/// will not uphold any security guarantees (i.e. they will break)
+pub fn add_le(x: &mut [u8], y: &[u8]) {
+    if x.len() != y.len() {
+        panic!("add_le: lengths of slices are not equal");
```
nvesely: `panic!` should only be used for unrecoverable errors. In general we should returning `Result`s in any functions where failure could be okay. We don't have any `Err` types yet, so just return `Result<(), ()>`.

Tom-Phinney: There is no reason to require that the user know the length of the nonce. Nonce incrementation by +1 is common. The requirement should be
```rust
if x.len() < y.len() {
+        panic!("add_le: length of increment exceeds length of nonce");
```

Tom-Phinney:
> Can you provide some examples of protocols that do this in the stream cipher context?

[IEC 62591](https://en.wikipedia.org/wiki/WirelessHART) and [IEC 62734](https://en.wikipedia.org/wiki/ISA100.11a), which are both used by heavy industry for wireless communications of sensors, valves, controllers, etc. in refineries, chemical plants, power plants (but not the power grid), pharmaceuticals, etc. Their nonce is constructed in part from a timeslot counter whose period exceeds the lifetime of the keys.

> One thing that's cool about stream ciphers is the ability to decrypt an arbitrary portion of a ciphertext. This is because they just generate a pseudorandom byte stream that is XORed with the plaintext, and no portion of this stream is dependent on any other. 

This is true for the stream ciphers that the NSA loves. It's not true for military stream ciphers, which tend to be autokeyed (i.e., using internal feedback to morph the key continuously) and are much harder to attack. One of the longest-suppressed patents in US history [was for such a stream cipher](http://patft.uspto.gov/netacgi/nph-Parser?Sect1=PTO1&Sect2=HITOFF&d=PALL&p=1&u=%2Fnetahtml%2FPTO%2Fsrchnum.htm&r=1&f=G&l=50&s1=6,912,284.PN.&OS=PN/6,912,284&RS=PN/6,912,284), filed in 1983 and issued in 2005 after one of my colleagues had filed a competing claim.

src/utils.rs
nvesely: I thought we all had just got on a tangent about Nonces. Apparently I glanced over both the initial PR comment and the docstring in the diff. The purpose of `sodium_add` is not to increment a nonce. See `sodium_increment` for a function that does that (in constant time). `sodium_add` performs little-endian modular addition on two large numbers represented as `char`s. There is actually no chance (barring implementation errors) of overflow because the addition is performed as `(a + b) mod 2^(8*len)` according to [the documentation](https://download.libsodium.org/doc/helpers/). So, this kind of all got off-track based on an initial misreading, but at least I think we've worked some things out for future reference.

Tom-Phinney: So if I want to increment the nonce by +1, I can use `sodium_increment`, but if I want to increment it by +2 I have to determine the nonce's length and create a bytes slice of equal length? Why is a shorter slice for the addend not acceptable? The only reason I can see is the potential for a side-channel attack on the duration of the `sodium_add` operation if the timing is variable, but that same attack applies to `sodium_increment`. If `sodium_increment` has constant timing, so can `sodium_add` with a user-selected input slice length for the increment.

kurnevsky:
> To address the discussion of Nonces, I'm against the user ever really dealing with them. This is something that should be abstracted away as much as possible because that's safer and there's not really a good reason afaict a user would need to deal with them at all. In my opinion, generally speaking a cryptography library should not even include a public nonce type. Instead, I would just transparently generate the nonce during encryption and encode it in the ciphertext.

Actually "inc"ing the nonce can be used to encrypt different messages when secure communication is established. And one practical example - in TOX protocol we establish connection over UDP and every sent message is encrypted with `nonce = base_nonce + message_number`. But this used nonce won't be included in the message - message includes only last two bytes of nonce (to reduce traffic usage and to increase probability that UDP packet will be delivered, I suppose). And since UDP doesn't guarantee that messages will be delivered in the same order they were sent we can't know the message number. So when we receive a message we take these two bytes, calculate the diff with two bytes of the base nonce we know and then "inc"ing the nonce with the diff.
So this involves a lot of manual work with nonces and we have to have the way to do it since it's part of the protocol.

src/utils.rs
kurnevsky: > See sodium_increment for a function that does that (in constant time). sodium_add performs little-endian modular addition on two large numbers represented as chars.

Right, `sodium_add` can be used to perform const-time addition of any byte strings whenever you need it. But it seems to me that the main case when you need it will be "inc"ing nonces by some value you know.

> The requirement should be `if x.len() < y.len() {`

`sodium_add` takes only one length argument for both byte strings. So if we allow to pass shorter `y` we will have to build a full string in this function before passing to libsodium. I'd rather keep it simple binding to `sodium_add`.

> panic! should only be used for unrecoverable errors.

I can change it to `Result` but in this case it's possible that user ignores the result and ends up with the same nonce (or whatever) he had before. And I'd say that we can panic not only on unrecoverable errors but in cases when we shouldn't deal with this kind of error at all. When we pass wrong arguments we shouldn't check afterwards if they were wrong and do something in this case.

Dylan-DPC: >  can change it to Result but in this case it's possible that user ignores the result and ends up with the same nonce (or whatever) he had before. 

Can you explain how a user will end up on the same nonce by ignoring the result? I can't think of any way this will happen.

Tom-Phinney:
Both of the above-cited IEC industrial comm protocols use a similar approach. In those protocols all members of a network share a synchronized sense of time, needed for synchronized frequency hopping and integrated send/receive comm scheduling of IEEE 802.15.4 radios. Network-wide multicast encryption at the Data Link layer isolates each network and limits propagation of outside DNS attacks, while point-to-point encryption at the Transport layer provides confidentiality and integrity. In both cases the nonces include the full synchronized sense of time, with only lsbs of that time-sense transported in the message headers.

src/utils.rs
Dylan-DPC: Yes but as crate maintainers, we can't decide on what the user might want to use. `Result` is a better way as the user can handle the error the way he needs or ignore it with `unwrap()` or something similar

kurnevsky: But instead of handling result he can just check lengths before calling this function.
Just let me know if it's already decided and I'll update the pr.

src/utils.rs
```diff
@@ -119,4 +137,59 @@ mod test {
             assert!(x.iter().all(|xi| *xi == 0));
         }
     }
+
+    #[test]
+    fn test_add_le_zero() {
+        for i in 1usize..256 {
```
nvesely: usize isn't necessary here

src/utils.rs
```diff
@@ -119,4 +137,59 @@ mod test {
...
+            let mut x = vec!(0u8; i);
+            let mut y = vec!(0u8; i);
```
nvesely: Doesn't need to be mutable

src/utils.rs
```diff
@@ -119,4 +137,59 @@ mod test {
...
+        assert_eq!(x, z);
+        let mut x = [255, 255, 255, 255, 255];
+        let z = [41, 0, 0, 0, 0];
+        add_le(&mut x, &y);
+        assert!(!x.iter().all(|x| { *x == 0 }));
+        assert_eq!(x, z);
+    }
+
+    #[test]
+    fn test_add_le_overflow() {
```
nvesely: What are you testing for here?

nvesely:
If there are ideas for how to make Nonce addition easier, or we want to implement additional traits for Nonce, then another issue should be opened to discuss that.

dnaq:
I’m against changing the function name to ‘sodium_add()’. That naming goes
against all naming conventions used in sodiumoxide. If the argument is that
using the same function names as in libsodium increases usability then I
would prefer each primitives documentation to link to the corresponding
documentation page of libsodium, not to use C-style naming conventions.

nvesely:
Cool with me @dnaq. That was the bullet I felt least strongly about. Can we all agree on the rest so @kurnevsky can have some clear direction?

dnaq:
I’m ok with the rest. Keep the warning though, I would prefer all nonce
handling functions to warn about nonce reuse.

nvesely:
I meant just get rid of the first line of the warning.

src/utils.rs
kurnevsky: The same thing that is tested [here](https://github.com/sodiumoxide/sodiumoxide/blob/master/src/utils.rs#L115) - that `add_le` works as modular addition.

src/utils.rs
kurnevsky: It does because of the next line. Or how else I can initialize a vector from all but one zeros?

kurnevsky:
It seems everything is resolved?

nvesely:
Almost.

> this method does not check for arithmetic overflow.

should still be removed.

nvesely:
You maybe want to use `memcmp` instead of `==`, but I'm not sure it matters here. Otherwise, LGTM.

kurnevsky:
If you are speaking about comparison of lengths then it's primitive type and `==` executes in constant time. And how do you imagine using `memcmp` here?

nvesely:
You're right. Not an array that's going to shortcircuit.

kurnevsky:
Fixed, thanks.

PR #214, opened 2018-05-23
fix std libraries being pulled in even in no_std environments (take 2)

### LINKED ISSUE #197: std is included even in no_std environments.
I am trying to write a piece of code targeting the x86_64 ABI using a custom target, which means there's no `std` available for my build target so I am building with `#[no_std]`. It seems that even with the std flag disabled, the package still tries to pull an `std` package into the build. The following is the build error:

```bash
$ xargo build --release --target=raw
   Compiling pkg-config v0.3.9
   Compiling libc v0.2.36
   Compiling rlibc v1.0.0
   Compiling compiler_builtins v0.1.0 (https://github.com/rust-lang-nursery/compiler-builtins#0ba07e49)
error[E0463]: can't find crate for `std`
  |
  = note: the `raw` target may not be installed

error: aborting due to previous error

```

Note I am not actually doing anything with the package yet, if I remove it from my dependencies the build succeeds and no `std` is ever built.

My environment:

#### Cargo.toml
```toml
# ...
[dependencies.sodiumoxide]                                                                                                                                                                                                                                                    
version = "*"                                                                                                                                                                                                                                                                 
default-features = false     
features=[]
```

#### raw.json
```json
  {                                                                                                                                                                                                                                                                             
    "pre-link-args"                    : ["-m64", "-nostdlib", "-static"],                                                                                                                                                                                                      
    "dynamic-linking"                  : false,                                                                                                                                                                                                                                 
    "relocation-model"                 : "pic",                                                                                                                                                                                                                                 
    "no-compiler-rt"                   : true,                                                                                                                                                                                                                                  
    "no-default-libraries"             : true,                                                                                                                                                                                                                                  
    "position-independent-executables" : false,                                                                                                                                                                                                                                 
    "llvm-target"                      : "x86_64-unknown-none",                                                                                                                                                                                                                 
    "data-layout"                      : "e-m:e-i64:64-f80:128-n8:16:32:64-S128",                                                                                                                                                                                               
    "linker-flavor"                    : "gcc",                                                                                                                                                                                                                                 
    "target-endian"                    : "little",                                                                                                                                                                                                                              
    "target-pointer-width"             : "64",                                                                                                                                                                                                                                  
    "target-c-int-width"               : "32",                                                                                                                                                                                                                                  
    "arch"                             : "x86_64",                                                                                                                                                                                                                              
    "os"                               : "none",                                                                                                                                                                                                                                
    "disable-redzone"                  : true,                                                                                                                                                                                                                                  
    "features"                         : "+avx,+mmx,+sse",                                                                                                                                                                                                                      
    "panic-strategy"                   : "abort"                                                                                                                                                                                                                                
  }
```

nvesely:
Supersedes #213 (I'll leave it up to @kinggoesgaming when/if to close that PR). Fixes #197.

Wanted to make some changes #213 since some of the dependencies in the manifest were still pulling in std.

@Reisen can you test this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,18 +16,18 @@ exclude = [
 ]
 
 [badges]
-travis-ci = { repository = "dnaq/sodiumoxide" }
+travis-ci = { repository = "sodiumoxide/sodiumoxide" }
 
 [dependencies]
-libc = "0.2"
+libc = { version = "^0.2.41" , default-features = false }
 libsodium-sys = { version = "0.0.16", path = "libsodium-sys" }
-serde = { version="1", optional = true }
+serde = { version = "^1.0.59", default-features = false, optional = true }
 
 [dev-dependencies]
-serde = "1"
-serde_json = "1"
-rustc-serialize = "0.3"
-rmp-serde = "0.13"
+serde = "^1.0.59"
+serde_json = "^1.0.17"
+rustc-serialize = "^0.3.24"
+rmp-serde = "^0.13.7"
 
 [features]
 benchmarks = []
--- a/libsodium-sys/Cargo.toml
+++ b/libsodium-sys/Cargo.toml
@@ -12,13 +12,13 @@ categories = ["cryptography", "api-bindings"]
 version = "0.0.16"
 
 [badges]
-travis-ci = { repository = "dnaq/sodiumoxide" }
+travis-ci = { repository = "sodiumoxide/sodiumoxide" }
 
 [build-dependencies]
-pkg-config = "0.3"
+pkg-config = "^0.3.11"
 
 [dependencies]
-libc = "0.2"
+libc = { version = "^0.2.41" , default-features = false }
 
 [lib]
 name = "libsodium_sys"
--- a/libsodium-sys/lib.rs
+++ b/libsodium-sys/lib.rs
@@ -1,4 +1,5 @@
 #![allow(non_upper_case_globals)]
+#![no_std]
 
 extern crate libc;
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -16,18 +16,18 @@ exclude = [
 ]
 
 [badges]
-travis-ci = { repository = "dnaq/sodiumoxide" }
+travis-ci = { repository = "sodiumoxide/sodiumoxide" }
 
 [dependencies]
-libc = "0.2"
+libc = { version = "^0.2.41" , default-features = false }
```
Dylan-DPC: you don't have to mention `^`
`x.y.z` = `^x.y.z`

nvesely: I didn't realize that was the default operator. Now I know.

I was reading https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html and I hadn't seen it before, but like the way the tilde operator works. I think that its use (i.e., specifying `"~x.y.z"`) could minimize breakage and also force us to be more intentional (e.g., try to review upstream changes) when upgrading minor versions of the (very few) dependencies of this crate. Thoughts?

Side note: an @alias for the sodiumoxide maintainers would be useful.

Dylan-DPC: ~ is good when you don't want to limit it to the a specific minor dependency, but I don't see the case for it here. most crates follow semver well enough to be backwards compatible. So x.y.z should be fine

nvesely: Fair enough. I'm going to leave this as-is then since the only "problem" is it's slightly verbose syntax.

Dylan-DPC:
bors: r+

PR #229, opened 2018-06-28
Minor test suite improvements

darrenldl:
Added quickcheck tests for `crypto/verify.rs`
  - `qc_verify_16`
  - `qc_verify_32`
  - `qc_verify_64`

Added 3 test vectors from other IETF documents for `crypto/aead/chacha20poly1305_ietf.rs`
  - `test_vector_2`
  - `test_vector_3`
  - `test_vector_4`

All tests passed.

Only test code is touched, except the `extern crate` part in `lib.rs` for using `quickcheck` during tests.


darrenldl:
Also I have no idea how to update brach properly - I recall a conversation some time ago that clicking "Update branch" button will mess up the commits(?).

Any advice on that would be great.

Dylan-DPC:
@darrenldl preferably rebase over `master`.

nvesely:
Looks good. What I do is set my local master to track `upstream/master` via the `git branch -u`. Then to create a PR `git checkout master && git pull && git checkout -b feature-branch-name`. To rebase a feature branch checkout and pull master again then run `git rebase master` with the feature branch checked out. 

Also when pushing you don’t want to push to your master. That’s confusing if you want to make more than one PR to a repo IMO and that’s why feature branches are useful. Use `git push -f origin HEAD`.

Hope you find that useful.

darrenldl:
@nvesely Thanks!

I had no idea you can do more than one PR at a time with feature branches, that will definitely speed up writing of the test suite.

I'll try that next time, and hope I get it right.

Dylan-DPC:
@darrenldl you can also do
```
git fetch origin
git rebase origin/master
```

(assuming origin points to this repo)

Dylan-DPC:
What's the gain of using `quickcheck` here?

darrenldl:
@Dylan-DPC I just tend to use it for "educated" random testing, cause it tends to find the breaking set of values very quickly, then shrink them effectively too. I'm using it as a poor man's fuzzing in this context basically.

For `verify_16`, `verify_32`, and `verify_64` specifically, it is basically to avoid having to try all combinations of differing configuration between two slices, while having a somewhat smart algorithm that can try to capture some corner cases. Though the `verify` functions are probably too simple to have any visible bugs, so this is largely just to establish a use case/template for later uses.

`proptest` would have worked better actually, but didn't think of it at the time as I'm more used to quickcheck.

Issue #201, opened 2018-04-08
Expose bindings to crypto_secretstream_*()


jrmuizel:
This is helpful for supporting data that's larger than will fit in memory

dnaq:
Please see #203

cynecx:
JFYI: I have a **experimental** branch which supports [this](https://github.com/cynecx/sodiumoxide/tree/streaming).

jrmuizel:
What needs to be done to get your branch mergeable?

Dylan-DPC:
@cynecx do you want to merge it? the longer you wait the more conflicts you will end up with

nvesely:
Let's see if @dkulic gets back this week, and if not go from there.

dkulic:
Hi, I am here. I will look for the conflicts.

nvesely:
There are two closed PRs that fix this issue: #222 and #240. #222 was closed because it seemed like there was momentum behind a re-implementation with a different API, but then too many people had problems with the new API implemented in #240 as well. There has been a ridiculous amount of back and forth over the issue, and not much progress has been made with respect to some sort of consensus on an interface.

Personally, I would like someone to implement https://github.com/sodiumoxide/sodiumoxide/pull/240#issuecomment-420765420, and I'd be happy to do it myself. I at least got two thumbs up emojis from core contributors to the project on this proposal, which kind of feels like enough, but due to the continued discussion, I didn't want to risk my time by implementing something that won't be merged and that will just spawn more debate/ discussion. I will only implement something I feel has broad consensus and that I feel is likely to be quickly merged with little modification.

My interest in further discussing interfaces here is pretty low; I've never seen such protracted discussion over such a small feature implementation. If people are not happy with my proposal, I'm totally okay with that, and am happy to let other people continue to work out this issue on their own

Dylan-DPC:
Yeah it is sad to be in a stalemate state where there is no clear advantage by one proposal over the other and there is no clear consensus. 

What we can do is "fastrack" this issue - that is whoever picks up just builds the complete feature without facing any questions (except for bugs during review process). 

If all fine, then any one (including @nvesely if they want) can pick this and go ahead and implement it without stoppages on the design choices. I know we might end up realising the flaws later and rewriting some of the stuff but i think its the best thing to do instead of moving in circles.

samscott89:
[...]

PR #240, opened 2018-07-13
Implement secretstream API

nvesely:
Supersedes #222. I couldn't request a review from you @dkulic, but please take a look.

### DIFF
--- a/src/crypto/secretstream/secretstream_macros.rs
+++ b/src/crypto/secretstream/secretstream_macros.rs
@@ -62,30 +62,30 @@ const TAG_FINAL: u8 = $tag_final as u8;
 pub enum Tag {
     /// Message, the most common tag, that doesn't add any information about the nature of the message.
     Message,
-
     /// Push: indicates that the message marks the end of a set of messages,
     /// but not the end of the stream.
     /// For example, a huge JSON string sent as multiple chunks can use this tag to indicate
     /// to the application that the string is complete and that it can be decoded.
     /// But the stream itself is not closed, and more data may follow.
     Push,
-
     /// Rekey: "forget" the key used to encrypt this message and the previous ones,
     /// and derive a new secret key.
     Rekey,
-
     /// Final: indicates that the message marks the end of the stream
     /// and erases the secret key used to encrypt the previous sequence.
     Final,
 }
 
-fn _tag_from_byte(tag: u8) -> Result<Tag, ()> {
-    match tag {
-        TAG_MESSAGE => Ok(Tag::Message),
-        TAG_PUSH => Ok(Tag::Push),
-        TAG_REKEY => Ok(Tag::Rekey),
-        TAG_FINAL => Ok(Tag::Final),
-        _ => Err(())
+impl Tag {
+    /// Returns the corresponding `Tag` given a `u8`, else `Err(())`.
+    fn from_u8(tag: u8) -> Result<Tag, ()> {
+        match tag {
+            TAG_MESSAGE => Ok(Tag::Message),
+            TAG_PUSH => Ok(Tag::Push),
+            TAG_REKEY => Ok(Tag::Rekey),
+            TAG_FINAL => Ok(Tag::Final),
+            _ => Err(())
+        }
     }
 }
 
@@ -137,7 +137,7 @@ impl Encryptor {
             return Err(());
         }
 
-        Ok((Self(state), Header(header)))
+        Ok((Encryptor(state), Header(header)))
     }
 
 
@@ -284,7 +284,7 @@ impl Decryptor {
             return Err(());
         }
 
-        let tag = _tag_from_byte(tag)?;
+        let tag = Tag::from_u8(tag)?;
         if tag == Tag::Final {
             self.flag_finalized = true;
         }
--- a/src/crypto/secretstream/secretstream_macros.rs
+++ b/src/crypto/secretstream/secretstream_macros.rs
@@ -17,6 +17,7 @@ macro_rules! stream_module (($state_name: ident,
 use libc::c_ulonglong;
 use randombytes::randombytes_into;
 use std::mem;
+use std::ptr;
 
 /// Returns the maximum length of an individual message.
 // TODO: use `const fn` when stable
@@ -132,12 +133,11 @@ impl Encryptor {
         let rc = unsafe {
             $init_push_name(&mut state, header.as_mut_ptr(), key.0.as_ptr())
         };
-
-        if rc == 0 {
-            Ok((Encryptor(state), Header(header)))
-        } else {
-            Err(())
+        if rc != 0 {
+            return Err(());
         }
+
+        Ok((Self(state), Header(header)))
     }
 
 
@@ -147,15 +147,14 @@ impl Encryptor {
     /// object.
     pub fn init_gen_key() -> Result<(Self, Header, Key), ()> {
         let key = gen_key();
-        let result = Self::init(&key);
 
-        if result.is_ok() {
-            let (encryptor, header) = result.unwrap();
-            Ok((encryptor, header, key))
-        } else {
-            Err(())
+        let result = Self::init(&key);
+        if result.is_err() {
+            return Err(());
         }
 
+        let (encryptor, header) = result.unwrap();
+        Ok((encryptor, header, key))
     }
 
     /// Encrypts a message `m` using the `state` and tags it as `Message`.
@@ -198,31 +197,30 @@ impl Encryptor {
     /// Additional data ad of length adlen can be included in the computation of the authentication tag.
     /// If no additional data is required, ad can be None.
     fn _push(&mut self, m: &[u8], ad: Option<&[u8]>, tag: u8) -> Result<Vec<u8>, ()> {
+        let mlen = m.len();
         if m.len() > messagebytes_max() {
             return Err(());
-        } else {
-            println!("{}", m.len());
-            println!("{}", messagebytes_max());
+        }
+        let clen = mlen + ABYTES;
+        let mut c = Vec::with_capacity(clen);
+        let (ad_p, ad_len) = ad.map(|ad| (ad.as_ptr(), ad.len()))
+                               .unwrap_or((ptr::null(), 0));
+
+        let rc = unsafe {
+            $push_name(&mut self.0,
+                       c.as_mut_ptr(),
+                       &mut (clen as c_ulonglong),
+                       m.as_ptr(),
+                       mlen as c_ulonglong,
+                       ad_p,
+                       ad_len as c_ulonglong,
+                       tag)
+        };
+        if rc != 0 {
+            return Err(());
         }
 
-        let (ad_p, ad_len) = ad.map(|ad| (ad.as_ptr(), ad.len() as c_ulonglong)).unwrap_or((0 as *const _, 0));
-        let mut c = Vec::with_capacity(m.len() + ABYTES);
-        let mut clen = c.capacity() as c_ulonglong;
-
-        unsafe {
-            let err = $push_name(&mut self.0,
-                                 c.as_mut_ptr(),
-                                 &mut clen,
-                                 m.as_ptr(),
-                                 m.len() as c_ulonglong,
-                                 ad_p,
-                                 ad_len,
-                                 tag);
-            if err != 0 {
-                return Err(());
-            }
-            c.set_len(clen as usize);
-        }
+        unsafe { c.set_len(clen) };
         Ok(c)
     }
 }
@@ -238,14 +236,17 @@ impl Decryptor {
     /// Initializes a `state` given a secret `key` and a `header`.
     /// The `key` k will not be required any more for subsequent operations.
     /// It returns Err if the header is invalid.
-    pub fn init(header: &Header, key: &Key) -> Result<Decryptor, ()> {
-        unsafe {
-            let mut state: $state_name = mem::uninitialized();
-            if $init_pull_name(&mut state, header.0.as_ptr(), key.0.as_ptr()) != 0 {
-                return Err(());
-            }
-            Ok(Decryptor{state, flag_finalized: false})
+    pub fn init(header: &Header, key: &Key) -> Result<Self, ()> {
+        let mut state: $state_name = unsafe { mem::uninitialized() };
+
+        let rc = unsafe {
+            $init_pull_name(&mut state, header.0.as_ptr(), key.0.as_ptr())
+        };
+        if rc != 0 {
+            return Err(());
         }
+
+        Ok(Self{state, flag_finalized: false})
     }
 
     /// Verifies that c (a sequence of bytes) contains a valid ciphertext and authentication tag
@@ -264,26 +265,32 @@ impl Decryptor {
         if mlen > messagebytes_max() {
             return Err(());
         }
-
         let mut m = Vec::with_capacity(mlen);
-        let (ad_p, ad_len) = ad.map(|ad| (ad.as_ptr(), ad.len() as c_ulonglong)).unwrap_or((0 as *const _, 0));
-        let mut tag: u8 = 0;
+        let (ad_p, ad_len) = ad.map(|ad| (ad.as_ptr(), ad.len()))
+                               .unwrap_or((ptr::null(), 0));
+        let mut tag: u8 = unsafe { mem::uninitialized() };
 
-        unsafe {
-            if $pull_name(&mut self.state,
-                          m.as_mut_ptr(),
-                          &mut (mlen as c_ulonglong),
-                          &mut tag,
-                          c.as_ptr(),
-                          clen as c_ulonglong,
-                          ad_p,
-                          ad_len) != 0 {
-                return Err(());
-            }
-            m.set_len(mlen as usize);
+        let rc = unsafe {
+            $pull_name(&mut self.state,
+                       m.as_mut_ptr(),
+                       &mut (mlen as c_ulonglong),
+                       &mut tag,
+                       c.as_ptr(),
+                       clen as c_ulonglong,
+                       ad_p,
+                       ad_len as c_ulonglong)
+        };
+        if rc != 0 {
+            return Err(());
         }
+
         let tag = _tag_from_byte(tag)?;
-        if tag == Tag::Final { self.flag_finalized = true; }
+        if tag == Tag::Final {
+            self.flag_finalized = true;
+        }
+
+        unsafe { m.set_len(mlen) }
+
         Ok((m, tag))
     }
 
--- a/src/crypto/secretstream/secretstream_macros.rs
+++ b/src/crypto/secretstream/secretstream_macros.rs
@@ -21,22 +21,64 @@ use std::mem;
 pub const KEYBYTES: usize = $keybytes as usize;
 
 /// Number of bytes in a `Header`.
+/// An encrypted stream starts with a short header, whose size is HEADERBYTES bytes.
+/// That header must be sent/stored before the sequence of encrypted messages,
+/// as it is required to decrypt the stream.
 pub const HEADERBYTES: usize = $headerbytes as usize;
 
-/// Number of added bytes
+/// Number of added bytes. The ciphertext length is guaranteed to always be message length + ABYTES.
 pub const ABYTES: usize = $abytes as usize;
 
-/// Tag message
-pub const TAG_MESSAGE: u8 = $tag_message as u8;
+/// Tag message, the most common tag, that doesn't add any information about the nature of the message.
+const TAG_MESSAGE: u8 = $tag_message as u8;
 
-/// Tag push
-pub const TAG_PUSH: u8 = $tag_push as u8;
+/// Tag push: indicates that the message marks the end of a set of messages,
+/// but not the end of the stream.
+/// For example, a huge JSON string sent as multiple chunks can use this tag to indicate
+/// to the application that the string is complete and that it can be decoded.
+/// But the stream itself is not closed, and more data may follow.
+const TAG_PUSH: u8 = $tag_push as u8;
 
-/// Tag rekey
-pub const TAG_REKEY: u8 = $tag_rekey as u8;
+/// Tag rekey: "forget" the key used to encrypt this message and the previous ones,
+/// and derive a new secret key.
+const TAG_REKEY: u8 = $tag_rekey as u8;
 
-/// Tag final
-pub const TAG_FINAL: u8 = $tag_final as u8;
+/// Tag final: indicates that the message marks the end of the stream
+/// and erases the secret key used to encrypt the previous sequence.
+const TAG_FINAL: u8 = $tag_final as u8;
+
+/// Tag of the message. When message is encrypted the tag is attached.
+/// When decrypting the tag is retrieved and may be used.
+#[derive(Debug, PartialEq)]
+pub enum Tag {
+    /// Message, the most common tag, that doesn't add any information about the nature of the message.
+    Message,
+
+    /// Push: indicates that the message marks the end of a set of messages,
+    /// but not the end of the stream.
+    /// For example, a huge JSON string sent as multiple chunks can use this tag to indicate
+    /// to the application that the string is complete and that it can be decoded.
+    /// But the stream itself is not closed, and more data may follow.
+    Push,
+
+    /// Rekey: "forget" the key used to encrypt this message and the previous ones,
+    /// and derive a new secret key.
+    Rekey,
+
+    /// Final: indicates that the message marks the end of the stream
+    /// and erases the secret key used to encrypt the previous sequence.
+    Final,
+}
+
+fn _tag_from_byte(tag: u8) -> Result<Tag, ()> {
+    match tag {
+        TAG_MESSAGE => Ok(Tag::Message),
+        TAG_PUSH => Ok(Tag::Push),
+        TAG_REKEY => Ok(Tag::Rekey),
+        TAG_FINAL => Ok(Tag::Final),
+        _ => Err(())
+    }
+}
 
 new_type! {
     /// `Key` for symmetric encryption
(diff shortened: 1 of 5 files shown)


src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,354 @@
+macro_rules! stream_module (($state_name: ident,
```
dnaq: I’m still not a fan of this being a macro. It seems premature to me, we have no idea if libsodium will add more primitives of this form, and even though it does we don’t know if the function signatures will be the same. I believe this shouldn’t be a macro for now, but might become one in the future if needed.

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,354 @@
...
+            TAG_PUSH => Ok(Tag::Push),
+            TAG_REKEY => Ok(Tag::Rekey),
+            TAG_FINAL => Ok(Tag::Final),
+            _ => Err(())
+        }
+    }
+}
+
+new_type! {
+    /// `Key` for symmetric encryption.
```
dnaq: Symmetric authenticated encryption. I think it’s important to write it out consequently throughout the documentation. This is not an encryption primitive, it’s an authenticated encryption primitive (with authenticated additional data)

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,354 @@
...
+        Ok((Encryptor(state), Header(header)))
+    }
+
+
+    /// Securely generates a key and uses it to initialize an `Encryptor`.
+    /// Returns the `Encryptor` object, a `Header` (which is needed by the
+    /// recipient to initialize a corresponding `Decryptor`), and the `Key`
+    /// object.
+    pub fn new() -> Result<(Self, Header, Key), ()> {
+        let key = Key::new();
```
dnaq: I think we might skip this methid completely. There is already a way to initialize a random key, and there is a way to create an Encryptor from a key, so this one feels superfluous.

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,354 @@
...
+            return Err(());
+        }
+
+        let (encryptor, header) = result.unwrap();
+        Ok((encryptor, header, key))
+    }
+
+    /// Encrypts a message `m` and its `tag`. Optionally attaches unencrypted
+    /// additional data `ad`. All data is authenticated.
+    fn aencrypt(&mut self, m: &[u8], ad: Option<&[u8]>, tag: u8) -> Result<Vec<u8>, ()> {
```
dnaq: Maybe we should call this ‘seal()’ instead?

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// decrypted out of order for this reason). Also may validate the optional
+    /// unencrypted additional data `ad` using the authentication tag attached to
+    /// `c`. Finally decrypts the ciphertext and tag, and checks the tag
+    /// validity.
+    /// If any authentication fails, the stream has already been finalized, or if
+    /// the tag byte for some reason does not correspond to a valid `Tag`,
+    /// returns `Err(())`. Otherwise returns the plaintext and the tag.
+    /// Applications will typically use a `while decryptor.is_not_finalized()`
+    /// loop to authenticate and decrypt a stream of messages.
+    pub fn vdecrypt(&mut self, c: &[u8], ad: Option<&[u8]>) -> Result<(Vec<u8>, Tag),()> {
```
dnaq: This could be called ‘open()’

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+}
+
+/// `Decryptor` contains the state for multi-part (streaming) computations. This
+/// allows the caller to process encryption of a sequence of multiple messages.
+/// After the last message of a valid stream containing the tag `Tag::Finalized`
+/// is decrypted, the `Decryptor` may no longer verify and decrypt messages or
+/// re-key itself.
+pub struct Decryptor {
```
dnaq: Same name discussion as above. Although I hate the name ‘Opener’, so I’m open to suggestions.

src/crypto/secretstream/secretstream_macros.rs
dnaq: Maybe we could consume self and return it here together with the tag, so that we don’t return self if the state is finalized. Then it won’t be possible for a user to try to keep using the state even though it has been finalized. That change would lead to the Tag enum having to contain the decrypter.

src/crypto/secretstream/secretstream_macros.rs
nvesely: Changed.

src/crypto/secretstream/secretstream_macros.rs
nvesely: Keeping as a convenience method.

nvesely:
I don't really see a problem keeping it as a macro. If there's some reasonable chance we'll want to convert it back to a macro in the future, may as well leave it.

I updated the docstrings to try to clarify.

I don't want to mess with the function names more. I think they're clear as is.

Rebased on master.

nvesely:
Okay... I'd be open to renaming the functions and structs if we can all agree on a set of names @sodiumoxide/owners. Maybe I was a little frustrated by how long this PR has been sitting when writing my last comment (which I take responsibility for in large part), and the late-process suggestion when the past PR thread was already so lengthy. I still think everything else should be merged as is, including leaving this code as a macro.

@dnaq suggested `Decryptor` -> `Opener`, `Decryptor::vdecrypt` -> `Opener::open` and `Encryptor::aencrypt{,_*}` -> `Encryptor::seal{,_*}`. In my opinion, this takes us further away from the cryptography, but perhaps makes clear the stream-like properties of the construction... perhaps. @dnaq was unsure of a new name for the `Encyptor`, so may I suggest `Signet`, the name for those rings used to seal envelopes, which I believe captures the spirit of their idea and also adds some fun struct names.

To make a slight improvement in terms of absolute cryptographic clarity, perhaps we could simply only rename `Decryptor` -> `VDecryptor`, which matches the already-existing-method `Decryptor::vdecrypt`.

Please share your thoughts and fresh ideas @sodiumoxide/owners :bulb:

cynecx:
I am not sure if this has been discussed before but I dont't think we should keep track of the "finalization" of the secretstream.

If I understand the internals of libsodium and docs correctly, it doesn't exactly forbid you to have multiple logical stream in one physical stream (secretstream), as in having multiple `crypto_secretstream_xchacha20poly1305_TAG_FINAL` tags. So in this regard, I don't think we should unnecessarily limit the use-case here (I've been working on a project which actually makes use of this).

Also is the separation of the libsodium api in two structs (`Opener` and `Decrypter`) really necessary? It seems rather higher-level than the usual apis in sodiumoxide. If we end up with unifying the impls we might end up with something more compact in my opinion (https://github.com/cynecx/sodiumoxide/commit/8d64b6e2450f070ea8f5fd97ba054731a28f9cb5).

dnaq:
To be honest I’m not really happy with the current secretstream api, so if
you have any ideas to simplify it I think we should be open to suggestions.
Could you sketch out your idea more in detail?

cynecx:
@dnaq The api I am currently using is pretty much this: https://github.com/cynecx/sodiumoxide/commit/8d64b6e2450f070ea8f5fd97ba054731a28f9cb5 (A further commit adds a getter to the secretstream header).

(sorry I am currently on my mobile, so I am unable to give you a tl;dr. I will try later)

dnaq:
@cynecx To be honest I like the API you posted a lot more than the one in this pull request.

With some simple type-level hackery we could get it to be a lot more type safe. First of all we always want newtypes for `Key`s and `Nonce`s, just like in the rest of sodiumoxide. The second trick is to do something like (writing from memory without a compiler, so might miss out on the syntax a bit):
```
struct Push;
struct Push;

struct State<T> { // omitted... }

impl<Push> State<Push> {
    fn init_push(...) -> State<Push> { // ... }
    fn push(...) -> ...;
}

impl<Pull> State<Pull> {
    fn init_pull(...) -> State<Pull> { // ... }
    fn pull(...) -> ...;
}

impl<T> State<T> {
    // common methods for push and pull
}
```

In this way we can get rid of the assert statements making sure that we're using pull/push mode, but without complicating the code/api.

cynecx:
@dnaq Actually I’ve just thought about that (It’s like the same pattern the byteorder crate uses with LittleEndian and BigEndian). I would really appreciate going in this direction for the api design. +1.

cynecx:
@dnaq I've had some spare time today and tried to adapt my branch accordingly, to what we've talked about. In my opinion, this api seems much more rusty than before: https://github.com/cynecx/sodiumoxide/commit/113a93786644266a77d06291fc9b418b198de124

dnaq:
That looks a lot nicer. I would rename ‘StateMode’ to mode, but the api
seems to be nice.

nvesely:
I like that version better too. When you feel it's ready @cynecx feel free to make a PR and I'll close this one. Then we can all do a proper review.

I had never seen 👻`PhantomData`👻 before, but it's a really cool struct. The only early feedback I wanted to provide is just relaying the information from [this subsection](https://doc.rust-lang.org/1.4.0/std/marker/struct.PhantomData.html#indicating-ownership) of the `PhantomData` docs:

> Adding a field of type PhantomData<T> also indicates that your struct owns data of type T. This in turn implies that when your struct is dropped, it may in turn drop one or more instances of the type T, though that may not be apparent from the other structure of the type itself. This is commonly necessary if the structure is using a raw pointer like *mut T whose referent may be dropped when the type is dropped, as a *mut T is otherwise not treated as owned.

> If your struct does not in fact own the data of type T, it is better to use a reference type, like PhantomData<&'a T> (ideally) or PhantomData<*const T> (if no lifetime applies), so as not to indicate ownership.

nvesely:
> @nvesely Perhaps you could take over the branch?

That I can do. Secretstream take 3!

> Just one small note about some weirdness I've just noticed. Regarding this: pub fn push(&mut self, tag: Tag, msg: &[u8], ad: Option<&[u8]>, out: &mut Vec<u8>), currently we are completely filling the out Vec with the cipher, but imo it does make more sense to append it to the buffer.

Not sure on what you mean by "append it to the buffer." In my/ @dkulic's version the function signature looks like `fn aencrypt(&mut self, m: &[u8], ad: Option<&[u8]>, tag: u8) -> Result<Vec<u8>, ()>` (https://github.com/sodiumoxide/sodiumoxide/pull/240/files#diff-5219d2be90ddb37cadd4e32af3527fffR177). Are you suggesting we pass in a stream-like object that supports like say anything that implements `std::fmt::Write`?

> I am not sure if this has been discussed before but I dont't think we should keep track of the "finalization" of the secretstream.
>
> If I understand the internals of libsodium and docs correctly, it doesn't exactly forbid you to have multiple logical stream in one physical stream (secretstream), as in having multiple crypto_secretstream_xchacha20poly1305_TAG_FINAL tags. So in this regard, I don't think we should unnecessarily limit the use cases here (I've been working on a project which actually makes use of this).

When would you use FINAL instead of PUSH? It seems like the situation you're describing you'd want to use PUSH and/or REKEY. Can you show me where/ how you're using it in your project? Also, what's a physical as opposed to logical stream here?

cynecx:
> That I can do. Secretstream take 3!

Thanks!

> Are you suggesting we pass in a stream-like object that supports like say anything that implements std::fmt::Write?

We could but I probably should have said "extend" instead of "append" (As in preserving the data existing in the given Vec). Right now the new-branch overwrites the given Vec but sometimes that might not be desirable. So instead of overwriting the Vec, we could just extend it with the cipher.

> When would you use FINAL instead of PUSH? It seems like the situation you're describing you'd want to use PUSH and/or REKEY. Can you show me where/ how you're using it in your project? Also, what's a physical as opposed to logical stream here?

The tags are just some kind of metadata so one can track the end of the stream (or multiple streams!), except for `crypto_secretstream_xchacha20poly1305_TAG_REKEY` which has actual side-effects.

For example, you could have something like that:

```
MESSAGE = 00
PUSH = 01
REKEY = 02
FINAL = PUSH | REKEY // One could say that FINAL is actually a PUSH with the little REKEY extra

<CIPHER> // This is the "physical stream"
    <MESSAGE> // |---
    <MESSAGE> // | one logical stream
    <FINAL>   // |---

    <MESSAGE> // |---
    <MESSAGE> // |
    <REKEY>   // | second logical stream
    <MESSAGE> // |
    <FINAL>   // |---
<CIPHER>

**Comment by nvesely** [comment]
I just remembered @jedisct1 chimed in on the issue of finalize before... https://github.com/sodiumoxide/sodiumoxide/pull/222#issuecomment-400950092. So I think you're misusing it @cynecx and we should consume `self` on `finalize`. I think the question is how should we do this.

Let's just talk about `State<Push>` first. We want to `drop` `self` after sending a message with the FINAL tag. While it's more minimal and close to upstream to have just the `push` method, that means the function signature needs to always take `mut self` and return `(Option<Self>, Result<(), ()>)`. This is because defensive programming dictates we check the error and return a result, and depending on whether the FINAL tag is sent we do or don't return `self`. Otherwise we can implement either separate methods for each tag, or just create `finalize` as a separate method and then the `push` method will return `Err(())` if it receives the FINAL tag or else proceed as normal. Either way all pushes always return just a `Result<(), ()>` (so the `finalize` method takes `mut self` and the rest, however implemented, take `&mut self`).

Next, let's consider `State<Pull>`. In this PR we are tracking whether we'd seen the FINAL flag in the `Decryptor` struct, but @cynecx was saying that they would like it if we consumed `self` instead. We need to report `_tag_from_byte` errors\*, so we can either decide that upon such an error we `drop` `self` as if it was a FINAL tag since the stream is corrupted (returning `Result<(Option<Self>, Tag), ()>`) or otherwise not `drop` `self` on an invalid tag error (returning `(Option<Self>, Result<Tag, ()>)`). Either way the signature is not as clean as when we track finalized in the struct and return just `Result<Tag, ()>`. Another thing I had in mind was the `while decryptor.is_not_finalized()` syntax in mind for looping over a stream. The cleanest thing I can think of if we're going to take `mut self` as an argument to `pull` is:

```rs
let mut decryptor = Some(State::<Pull>::init(&key, &header)?);
while decryptor.is_some() {
    decryptor, tag = decryptor.pull(&c, &ad, &mut out)?;
    // Do stuff
}
```

^ At least I think that would work. Maybe otherwise:

```rs
let mut decryptor = State::<Pull>::init(&key, &header)?;
loop {
    decryptor, tag = decryptor.pull(&c, &ad, &mut out)?;
    // Do stuff
    if decryptor.is_none {
        break
    }
}
```
Compare that to how it works when tracking finalized internally:

```rs
let mut decryptor = State::<Pull>::init(&key, &header)?;
while decryptor.is_not_finalized() {
    tag = decryptor.pull(&c, &ad, &mut out)?;
    // Do stuff
}
```

If you have a better idea of how any of this could be done @cynecx @dnaq, please let me know. Anyway, recapping the above and making some decisions on alternatives I presented, my choices as of now would be:

(i) Add a `finalized: Option<bool>` field to `State`, which is only `Some` for `State<Pull>`.
(ii) Add a `finalize` method to `State<Push>` which takes `mut self` and for all other tags the programmer should use `push_*` methods defined for each tag. Trying to group all the other tags into a single function means checking the FINAL tag is not passed, which won't be a compile-time check. And creating a second enum that includes all tags, but FINAL would be ugly.
(iii) Keep `pull` as-is in this PR and also implement `is{,_not}_finalized` methods for `State<Pull>`.

\* Note to self to consider implementing the `TryFrom<u8>` trait for `Tag` behind a "nightly" feature gate.

samscott89:
Would it make sense to make the APIs more inline with the rest of sodiumoxide? I.e:
```rust
// (inside mod secretstream)
pub use xchacha20poly1305::*;
enum Push {}
enum Pull {}
pub trait Mode {}
impl Mode for Push {}
impl Mode for Pull {}
mod xchacha20poly1305 {
    use super::*;

    pub fn init<Mode>(header: &Header, key: &Key) -> State<Mode> {
        ...
    }

    pub fn push(state: &State<Push>, msg: &[u8], ad: Option<&[u8]>) -> Result<(), ()> {
        ...
    }
}
```

and so on? It mirrors quite nicely (imo) the API elsewhere:

```rust
use sodiumoxide::crypto::secretstream::{self, Mode, Push, Pull};

let key = secretstream::gen_key();
let state = secretstream::init::<Push>(&key);
let plaintext = b"some data";
let ad = b"Some additional data";
let ciphertext = secretstream::push(plaintext, Some(ad), &mut state);
 ...

let their_state = secretstream::init::<Pull>(&key);
let their_plaintext = secretstream::pull(&ciphertext, Some(ad), &mut state).unwrap();
assert!(plaintext == &their_plaintext[..]);
```
^^^ This is virtually copied from the other examples. It's basically `s/&key/&mut state/` and that's it.

dnaq:
@samscott89 I see your point, but we also have stateful APIs in sodiumoxide that are implemented as methods on a state-struct, e.g. `crypto::hash::sha{256,512}::State` and `crypto::auth::*::State`, so I think that would be a better fit for us.

samscott89:
Yeah, that's true. My main motivation for suggesting it was to make the types/functionality a bit more transparent. `let state = State::init(...);` followed by `state.push(msg)` doesn't (to me) convey the intent accurately, which is kind of shown by needing to call the variable something else `encryptor = State::init(...)` or whatever. I can see it with hashing, the idea of updating and finalizing is better established. Personally I like the explicitness of `secretstream::push`.

Hopefully this isn't too bikesheddy, I think the overall API design is looking good, it certainly felt natural to write out the above examples. Just wanted to throw some ideas in the ring.

samscott89:
Another small win for the module method approach, nicer errors:
```rust
   |
68 |     let ct = secretstream::push(&[1, 2, 3], &mut state);
   |                                             ^^^^^^^^^^ expected enum `secretstream::Push`, found enum `secretstream::Pull`
   |
   = note: expected type `&mut secretstream::State<secretstream::Push>`
              found type `&mut secretstream::State<secretstream::Pull>`
```

vs

```
42 |     pub struct State<T: Mode> {
   |     ------------------------- method `push` not found for this
...
67 |     let ct = state.push(&[1, 2, 3]);
```

nvesely:
I agree with @dnaq that we should use a stateful API here. To me @samscott89's version just feels more clunky (both for us writing it and the programmer using it or reading our code) and not like the standard OOP interfaces I'm used to.

Final suggestion (I hope!) for `State`. This is building off my last comment and the code in the branch @cynecx linked:

```rust
pub struct State<T: StateMode> {
    state: crypto_secretstream_xchacha20poly1305_state,
    finalized: bool,
    phantom: PhantomData<&T>,
}
```

First, I want to note I got rid of the `header` field as in @cynecx's code. Putting it in `State` didn't really make sense to me and I feel like increases the risk a programmer serializes/ saves `State`, which contains the `state` they should dispose of, when they really only want to save the `header`.

The reason I want to track `finalized` for both is that we had this goal of consuming `self` once finalized (hence not needing to track it), but I think that creates its own set of problems that's worse and nobody has come up with a solution to. First is that it requires either (i) separate `push_*` methods for each `Tag` or (ii) at least a separate `finalize` method and we lose compile-time checking on `Tag` (see (ii) from my last comment) or (iii) that we keep a single `push` and return `Option<self>`, which is annoying. By tracking `finalized` we can keep a single `push` method that returns an error if it's already finalized. Likewise with `pull` as I showed above, not only is it cumbersome to return `Option<self>` each time, but it doesn't lend itself nicely to looping over an incoming stream-like object like the `while state.is_not_finalized()` syntax does.

Let's address the reasons I think we wanted to consume `self`. First, to ensure the sensitive state material is wiped as quickly as possible, and second, to signal unambiguously to the programmer that it's all over once `finalize` is called or the FINAL tag is received. In the first case, what we can do is include `if` blocks in `push` and `pull` that not only set the `self.finalized` flag, but overwrite the `self.state` field (although I think this may be effectively done upstream anyway by an explicit rekeying, so maybe we don't need to, at least for push*). As to the second, the programmer will receive an `Err(())` if they try to keep operating on a `State` object after finalization and while unfortunately all our errors are generic here, it should still get them to investigate. Explaining clearly in our documentation that if you continue to operate on `State` after finalization you will receive errors will help, as will the `while state.is_not_finalized()` syntax which we can feature in example code snippets in our top-level module documentation.

I think the idea that tracking `finalized` because upstream doesn't is bad is wrong, and we can think of it as an extension of `state` that's just as invisible to the user. I think this version would be Rusty in its use of generics to enforce safety while closely mirroring the upstream API.

\* To be looked into further and determined upon implementation.

nvesely:
@dnaq @samscott89 @Dylan-DPC @Tom-Phinney @kpcyrd @cynecx I want to get us all on the same page with this PR. I have reached out to some of you via Gitter and not received a reply, so I'm trying here now. I would like to re-work this using the interface described in my most recent previous comment. I got a "thumbs up" emoji from @Dylan-DPC, but otherwise have not heard feedback.

We've already thrown a whole lot of ideas out there regarding interface, and I tried to make something that incorporated many ideas and concerns. I would really like to just get on implementing _something_, but want to make sure I can at least get two approvals, and no strong objections, before proceeding.

dnaq:
Ok, so this is what I would like to see from the interface.

`State<Pull>` and `State<Push>`.

`State<Pull>` should have method `fn init_pull(...) -> Self`, `State<Push>` has the corresponding `fn int_push(...) -> Self`. This makes the ergonomics so that a caller can call `State::init_push()` or `State::init_pull()` without further type-annotations, which should make @samscott89 happy.... :)

`State<Push>` has the method `push` that takes a `PushTag` (name open for discussion), which doesn't include `Final`, to send the final tag we can call `finalize(mut self)` which consumes the state.

`State<Pull>` has the method `pull` that in addition to the decrypted/authed data returns a `PullTag` (name open for discussion) that includes `Final`. If final is received it should not be able to call it again, but due to ergonomics `pull` shouldn't consume `self`.

Both types of states should expose a `rekey`-method (could be defined on `impl<T> State<T> ...` that does a rekey without sending a suitable tag. That method should be documented, as in the libsodium API, so that users don't try to use it unnecessarily (most users should rekey by giving a `PushTag::Rekey` to `push`, causing the corresponding `pull` to rekey automatically). 

We also need to add the test vectors from libsodium, and add tests for all combinations of `Tag::Push`, `Tag::Message`, `Tag::Rekey` and `Tag::Final`.

Anything I've missed?

dkulic:
Hi guys,

Sorry I have been absent for a long time.

@dnaq may I implement you proposal?

dnaq:
@dkulic Sure, would be happy to see it implemented.

samscott89:
> State<Pull> should have method fn init_pull(...) -> Self, State<Push> has the corresponding fn int_push(...) -> Self. This makes the ergonomics so that a caller can call State::init_push() or State::init_pull() without further type-annotations, which should make @samscott89 happy.... :)

Oho, you underestimate me :)
But really, my main complaint is to do with the semantics of "pushing a state" in order to encrypt a message. On gitter, I wrote:

My only issue with [this pr] is still that `let mut decryptor = State::<Pull>::init(&key, &header)?;` feels a little wonky, mostly due to this implication that `State<Pull>` is a `decryptor`. My latest suggestion was to make it syntactically `secretstream::push`, but I agree that keeping it OOP-based is probably nicer. In which case, could we consider just renaming `State` to something like `Stream`?
As a further possibility, could even make it `Stream<Send>` and `Stream<Recv>` which is even clearer, but perhaps straying too far from the original methods?

dnaq:
I'm ok with calling it `Stream`, or even `SecretStream`. Might be a bit redundant to write out `secretstream::Stream::init_push()` instead of `secretstream::State::init_push()`, but that's not really a huge issue. Calling it `State` would keep the same naming convention as we have in all other stream APIs though, so on further thought I think I prefer the name `State`.

I'm not sure about changing the names of `Push` and `Pull`, even though I'm not a big fan of the names (especially since in libsodium we can `push()` using a `TAG_PUSH`, that is the same name is used for two completely different things in the same api).

samscott89:
> I'm not sure about changing the names of Push and Pull, even though I'm not a big fan of the names (especially since in libsodium we can push() using a TAG_PUSH, that is the same name is used for two completely different things in the same api).

Yeah agree it's probably not worth the confusion in changing names. It's clear enough as it is.

> Calling it State would keep the same naming convention as we have in all other stream APIs though, so on further thought I think I prefer the name State.

This is my main disagreement. IMO streaming encryption is closer to encryption than to hashing/message authentication tags. So we should be producing an api closer to encryption. Seems like I'm in the minority in this though, so I'm not going to belabour the point any more.
My preferred would probably look like `let mut stream = secretstream::init_push();` followed by  `stream.push(&msg)` where `stream` is `SecretStream<Push>` and follows @nvesely's proposed API with tracking `finalized` etc.

nvesely:
Closing because I am no longer trying to merge this PR and discussion should be moved to #201.

jedisct1:
I would still really love to know what could have been done differently in the upstream library in order to make these bindings easier to design.

What are the root causes of all these discussions that eventually didn't achieve a consensus?

Are the tags confusing? Or rekeying? Or the semantics of the `final` tag?

dnaq:
@jedisct1, I believe the API is quite nice as a C api. In C it’s usually quite ok to require the caller to keep track of the underlying state, e.g. is it push/pull and so on.

There was some discussion on the final-tag, but that was more of a documentation issue.

The only (very minor) issue I have with the C API is the naming of the tags, e.g. TAG_PUSH means one thing, init_push means a completely different thing. That’s not something that has any bearing on this issue though.

It has been a bit hard to translate into idiomatic rust though.

Issue #226, opened 2018-06-21
Cargo doc breaking builds


Dylan-DPC:
This is a tracking issue for [serde doc issue](https://github.com/serde-rs/serde/issues/1318) which is currently breaking our [travis builds](https://travis-ci.org/sodiumoxide/sodiumoxide/jobs/395074939). 

This has resulted in a [compiler issue](https://github.com/rust-lang/rust/issues/51661). There is nothing to be done on our side, so this issue will be closed once rust team resolves it.

PR #259, opened 2018-10-03
Removes duct-taped allow failures

Dylan-DPC:
We had added allow-failures because of #226. Removing it to revert it back to the old config. 

Resolves: #226


Dylan-DPC:
Have bypassed the review permissions just for this PR since it is related to travis. 

bors: r+

PR #276, opened 2018-11-13
Fix build on Win using vcpkg

kpp:
1. Don't call to pkg-config or vcpkg if 1 of SODIUM_LIB_DIR or SODIUM_INC_DIR is set
2. Fix redundant linker flag specified for library `sodium` (This time for Win 10)
3. Add VCPKGRS_DYNAMIC to rerun-if-env-changed on Win systems, because vcpkg will fail if there is no such env var
4. Print full error message instead of "Could not find libsodium on this system"


Dylan-DPC:
bors: r+ p=10

kpp:
bors: retry

Issue #231, opened 2018-06-29
Pin version of libsodium


nvesely:
Each version of sodiumoxide should be pinned to a specific version of libsodium. A `build.rs` should fail to compile with an informative error message if linked the wrong version of libsodium. An idea is to create `libsodium-version.txt` in the top-level directory that just contains the libsodium version string such that `build.rs`, Travis, AppVeyor, etc. can all read it, making updating easier and reducing the chance of diversion (though we should notice that immediately because the build will fail anyway).

nvesely:
Btw we're testing 1.0.15 in Travis whereas 1.0.16 is the latest version.

dnaq:
Why pin to a specific version? Libsodium follows semver, so pin to a
minimum version instead.

kpcyrd:
Please note that pinning a specific version could cause issues for linux distributions. Requiring a minimum version to build successfully seems more sensible, but the build is already failing in that case anyway.

nvesely:
> Please note that pinning a specific version could cause issues for linux distributions. Requiring a minimum version to build successfully seems more sensible, but the build is already failing in that case anyway.

See https://github.com/sodiumoxide/sodiumoxide/issues/234#issuecomment-414797999. I think we can add this without causing any issues for any platform.

PR #274, opened 2018-11-09
Fix redundant linker flag specified for library `sodium`

kpp:
Close #273 

```
$ cargo build
   Compiling libsodium-sys v0.1.0 (/home/USER/sodiumoxide/libsodium-sys)
   Compiling sodiumoxide v0.1.0 (/home/USER/sodiumoxide)
warning: unused macro definition
   --> src/crypto/onetimeauth/../auth/auth_macros.rs:172:1
    |
172 | / macro_rules! auth_state (($state_name:ident,
173 | |                           $init_name:ident,
174 | |                           $update_name:ident,
175 | |                           $final_name:ident,
...   |
268 | | }
269 | | ));
    | |___^
    |
    = note: #[warn(unused_macros)] on by default

    Finished dev [unoptimized + debuginfo] target(s) in 7.27s
```

Issue #267, opened 2018-10-25
libsodium-sys does not compile


AljoschaMeyer:
Running `cargo check` for a project depending on  sodiumoxide 0.1.0 results in the following error message:

```
    Checking libsodium-sys v0.1.0                                                                                           
error: expected one of `crate`, `fn`, `pub`, `static`, `type`, or `}`, found `;`                                            
   --> /home/aljoscha/projects/rsssb/ssb-multiformats/target/debug/build/libsodium-sys-20b3a7a6f7fc0b79/out/sodium_bindings.rs:378:5
    |                                                                                                                       
377 | extern "C" {                                                                                                          
    |             - expected one of `crate`, `fn`, `pub`, `static`, `type`, or `}` here                                     
378 |     ;                                                                                                                 
    |     ^ unexpected token                                                                                                
                                                                                                                            
error: aborting due to previous error                                                                                       
                                                                                                                            
error: Could not compile `libsodium-sys`.
```

The file in question has a bunch of the following blocks:

```rust
extern "C" {
    ;
}
```

`cargo clean` doesn't change anything, it recompiles dependencies and gives up once it reaches libsodium-sys.

`rustc --version --verbose`:

```
rustc 1.31.0-nightly (f99911a4a 2018-10-23)
binary: rustc
commit-hash: f99911a4a0bead7dd1f9ef2f90442844434cc391
commit-date: 2018-10-23
host: x86_64-unknown-linux-gnu
release: 1.31.0-nightly
LLVM version: 8.0
```

kpcyrd:
Could you also provide the libsodium version installed on your system and which distro you're using?

AljoschaMeyer:
Distro: arch linux

`uname -a`: `Linux aljoscha-laptop 4.18.16-arch1-1-ARCH #1 SMP PREEMPT Sat Oct 20 22:06:45 UTC 2018 x86_64 GNU/Linux`

`pacman -Q libsodium`: `libsodium 1.0.16-1`

https://www.archlinux.org/packages/community/x86_64/libsodium/

kpp:
@AljoschaMeyer would you please try to run `cargo build` again on fresh `master`?

kpp:
@AljoschaMeyer would you please install `cargo-outdated` and give me the output of `cargo outdated`?

AljoschaMeyer:
- `git clone https://github.com/sodiumoxide/sodiumoxide.git`
- `cd sodiumoxide`
[...]

PR #279, opened 2018-11-13
#234 - option to download / compile libsodium

neonphog:
Pull code from rust_sodium-sys build.rs to address [#234](https://github.com/sodiumoxide/sodiumoxide/issues/234).

- Will use system library by default
- If a system library is not found, will download / compile
- Introduces `SODIUM_FROM_SRC` to force download / compile

In the interest of doing one thing at a time, does not check for incompatible system lib version (https://github.com/sodiumoxide/sodiumoxide/issues/231).

### DIFF
--- a/libsodium-sys/build.rs
+++ b/libsodium-sys/build.rs
@@ -1,18 +1,57 @@
 extern crate bindgen;
+#[cfg(not(windows))]
+extern crate cc;
+#[cfg(not(target_env = "msvc"))]
+extern crate flate2;
+extern crate http_req;
+#[cfg(target_env = "msvc")]
+extern crate libc;
 extern crate pkg_config;
-
-use std::env;
-use std::path::PathBuf;
-
+extern crate sha2;
+#[cfg(not(target_env = "msvc"))]
+extern crate tar;
+#[macro_use]
+extern crate unwrap;
 #[cfg(target_env = "msvc")]
 extern crate vcpkg;
+#[cfg(target_env = "msvc")]
+extern crate zip;
+
+use http_req::request;
+use sha2::{Digest, Sha256};
+use std::env;
+use std::fs;
+use std::io::Cursor;
+use std::path::{Path, PathBuf};
+
+static DOWNLOAD_BASE_URL: &'static str = "https://download.libsodium.org/libsodium/releases/";
+static VERSION: &'static str = "1.0.16";
+
+#[cfg(target_env = "msvc")] // libsodium-<VERSION>-msvc.zip
+static SHA256: &'static str = "0580d54f57594a7cb493607cec6e7045369fb67d43623491523781e901589948";
+
+#[cfg(all(windows, not(target_env = "msvc")))] // libsodium-<VERSION>-mingw.tar.gz
+static SHA256: &'static str = "5b81a4fc5d0de36dbda7efeaf355c133d4f6cc0b4dbf69bbe46ef7f5a6baa639";
+
+#[cfg(not(windows))] // libsodium-<VERSION>.tar.gz
+static SHA256: &'static str = "eeadc7e1e1bcef09680fb4837d448fbdf57224978f865ac1c16745868fbd0533";
 
 fn main() {
     println!("cargo:rerun-if-env-changed=SODIUM_LIB_DIR");
     println!("cargo:rerun-if-env-changed=SODIUM_INC_DIR");
     println!("cargo:rerun-if-env-changed=SODIUM_STATIC");
+    println!("cargo:rerun-if-env-changed=SODIUM_FROM_SRC");
 
-    let include_dir = find_libsodium();
+    let include_dir = {
+        if env::var("SODIUM_FROM_SRC").is_ok() {
+            get_libsodium()
+        } else {
+            match find_libsodium() {
+                Some(lib) => lib,
+                None => get_libsodium(),
+            }
+        }
+    };
 
     let bindings = bindgen::Builder::default()
         .header("sodium_wrapper.h")
@@ -30,47 +69,48 @@ fn main() {
         .expect("Couldn't write bindings!");
 }
 
+fn print_link(name: &str) {
+    let mode = match env::var_os("SODIUM_STATIC") {
+        Some(_) => "static",
+        None => "dylib",
+    };
+    println!("cargo:rustc-link-lib={}={}", mode, name);
+}
+
 #[cfg(target_env = "msvc")]
-fn find_libsodium() -> String {
+fn find_libsodium() -> Option<String> {
     if let Ok(lib_dir) = env::var("SODIUM_LIB_DIR") {
-        let mode = match env::var_os("SODIUM_STATIC") {
-            Some(_) => "static",
-            None => "dylib",
-        };
-        println!("cargo:rustc-link-lib={0}=libsodium", mode);
+        print_link("libsodium");
         println!("cargo:rustc-link-search=native={}", lib_dir);
     } else {
         if !try_vcpkg() {
-            panic!("Could not find libsodium on this system!")
+            return None;
+        }
+    }
+
+    let include_dir = env::var("SODIUM_INC_DIR")
+        .ok()
+        .or_else(get_vcpkg_include_path);
+
+    Some(include_dir)
+}
+
+#[cfg(not(target_env = "msvc"))]
+fn find_libsodium() -> Option<String> {
+    if let Ok(lib_dir) = env::var("SODIUM_LIB_DIR") {
+        print_link("sodium");
+        println!("cargo:rustc-link-search=native={}", lib_dir);
+    } else {
+        let statik = env::var_os("SODIUM_STATIC").is_some();
+        if let Err(_) = pkg_config::Config::new().statik(statik).find("libsodium") {
+            return None;
         }
     }
 
     let include_dir =
-        env::var("SODIUM_INC_DIR").ok()
-        .or_else(get_vcpkg_include_path).unwrap();
+        env::var("SODIUM_INC_DIR").or_else(|_| pkg_config::get_variable("libsodium", "includedir"));
 
-    include_dir
-}
-
-#[cfg(not(target_env = "msvc"))]
-fn find_libsodium() -> String {
-    if let Ok(lib_dir) = env::var("SODIUM_LIB_DIR") {
-        let mode = match env::var_os("SODIUM_STATIC") {
-            Some(_) => "static",
-            None => "dylib",
-        };
-        println!("cargo:rustc-link-lib={0}=sodium", mode);
-        println!("cargo:rustc-link-search=native={}", lib_dir);
-    } else {
-        let statik = env::var_os("SODIUM_STATIC").is_some();
-        pkg_config::Config::new().statik(statik).find("libsodium").unwrap();
-    }
-
-    let include_dir =
-        env::var("SODIUM_INC_DIR")
-        .or_else(|_| pkg_config::get_variable("libsodium", "includedir")).unwrap();
-
-    include_dir
+    include_dir.ok()
 }
 
 #[cfg(target_env = "msvc")]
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
@kpp can you check why the [appveyor build](https://ci.appveyor.com/project/Dylan-DPC/sodiumoxide/builds/20272968/job/rp00qofjoqk2he6s) is failing?

Dylan-DPC:
bors: try

libsodium-sys/build.rs
```diff
@@ -42,55 +94,406 @@ fn main() {
...
 }
 
 /* Must be called when no SODIUM_LIB_DIR and no SODIUM_INC_DIR env vars are set
 This function will set `cargo` flags.
 Return: the first include path from vcpkg
 */
 #[cfg(target_env = "msvc")]
-fn find_libsodium_pkg() -> String {
-    let lib = vcpkg::probe_package("libsodium")
-        .expect("Could not find libsodium on this system");
```
kpp: I personally prefer a strict mode.

neonphog: @kpp - to clarify, are you suggesting a mode where if the lib is not found on the system, it errors rather than downloading / compiling? Something like `SODIUM_REQUIRE_SYSTEM`?

kpp: Something like "Don't break the existing process". A user may want to see errors why the system library was not found especially when they set `SODIUM_LIB_DIR` and `SODIUM_INC_DIR`.

I suggest to build from src only when there are no env `SODIUM_LIB_DIR` and `SODIUM_INC_DIR` variables, and when there is `SODIUM_FROM_SRC` is set. We should report an error when there are `SODIUM_LIB_DIR` and `SODIUM_INC_DIR` and `SODIUM_FROM_SRC` are set.

neonphog: hmm. I think that makes sense for when `SODIUM_LIB_DIR` or `SODIUM_INC_DIR` are set, and can make that change.

What do you think about when there are no environment variables? My main hope is that this crate can be usable in a majority of environments without any special configuration... which probably means building from source, especially once https://github.com/sodiumoxide/sodiumoxide/issues/231 is included.

kpp: If there are no env variables we should do everything to make `sodiumoxide` compile, first search for a system library then try to build it from source. The main question is whether we should build libsodium static in this case because if there is no system library installed the user will get a missing .so when they try to run binary.

libsodium-sys/build.rs
```diff
@@ -42,55 +94,406 @@ fn main() {
...
+    println!(
+        "cargo:rustc-link-search=native={}",
+        lib_install_dir.display()
+    );
+
+    Some(format!("{}/include", install_dir))
+}
+
+#[cfg(not(windows))]
+fn build_libsodium() -> Option<String> {
```
kurnevsky: It seems `build_libsodium` can't return `None`? If so then maybe return just `String`?

neonphog: updated :+1:

libsodium-sys/build.rs
neonphog: @kpp - updated per this conversation.
- if `SODUIM_LIB_DIR` or `SODIUM_INC_DIR` are specified, will not build from source, and will error if `SODIUM_FROM_SRC` is specified
- will build from source if nothing is specified but a system lib is not found

libsodium-sys/build.rs
```diff
@@ -1,38 +1,91 @@
...
-        let inc_dir_isset = env::var_os("SODIUM_INC_DIR").is_some();
-        if lib_dir_isset || inc_dir_isset {
-            find_libsodium_env()
+        if from_src {
+            Some(build_libsodium())
         } else {
-            find_libsodium_pkg()
+            let lib_dir_isset = env::var_os("SODIUM_LIB_DIR").is_some();
+            let inc_dir_isset = env::var_os("SODIUM_INC_DIR").is_some();
+            if lib_dir_isset || inc_dir_isset {
```
kpp: This is gonna always be false since we are in `if from src {} else { HERE }`

libsodium-sys/build.rs
```diff
@@ -1,38 +1,91 @@
...
+                if from_src {
```
kpp: This is gonna always be false since we are in `if from src {} else { HERE }`

libsodium-sys/build.rs
```diff
@@ -42,55 +95,406 @@ fn main() {
...
 
 /* Must be called when no SODIUM_LIB_DIR and no SODIUM_INC_DIR env vars are set
 This function will set `cargo` flags.
 Return: includedir from pkg-config
 */
 #[cfg(not(target_env = "msvc"))]
-fn find_libsodium_pkg() -> String {
+fn find_libsodium_pkg() -> Option<String> {
     let statik = env::var_os("SODIUM_STATIC").is_some();
-    pkg_config::Config::new().statik(statik).probe("libsodium")
```
kpp: Still you can use `cargo:warning=TEXT` to notify the user if smth goes wrong

libsodium-sys/build.rs
```diff
@@ -1,38 +1,91 @@
...
+    }
+
     let bindings = bindgen::Builder::default()
         .header("sodium_wrapper.h")
         .whitelist_function("(sodium|crypto|randombytes)_.*")
         .whitelist_type("(sodium|crypto|randombytes)_.*")
         .whitelist_var("(sodium|crypto|randombytes)_.*")
         .clang_arg("-isystem")
-        .clang_arg(include_dir)
+        .clang_arg(include_dir.unwrap())
```
kpp: @Dylan-DPC @kurnevsky and I came to a conclusion that we must get rid of bindgen dependency since libsodium .h files do not have any platform-dependent code.

Since there are issues like that https://github.com/sodiumoxide/sodiumoxide/issues/280 and https://github.com/sodiumoxide/sodiumoxide/issues/267 and https://github.com/sodiumoxide/sodiumoxide/issues/231 the best option is to use the same scenario as rust_sodium does: https://github.com/maidsafe/rust_sodium/blob/master/rust_sodium-sys/build.rs#L64

libsodium-sys/build.rs
```diff
@@ -1,38 +1,91 @@
...
         }
     };
 
+    let mode = match env::var_os("SODIUM_STATIC") {
+        Some(_) => "static",
+        None => "dylib",
+    };
+
+    if cfg!(target_env = "msvc") {
+        println!("cargo:rustc-link-lib={0}=libsodium", mode);
```
kpp: `vcpkg::probe_package` and `pkg_config::Config::new().statik(statik).probe` already print these flags so you will get these warning during compilation: https://github.com/sodiumoxide/sodiumoxide/pull/274

libsodium-sys/build.rs
```diff
@@ -42,55 +95,406 @@ fn main() {
...
+        println!(
+            "cargo:warning=The path to the usual build directory contains spaces and hence \
+             can't be used to build libsodium.  Falling back to use {}.  If running `cargo \
+             clean`, ensure you also delete this fallback directory",
+            fallback_path
+        );
+    }
+
+    // Create directories
+    unwrap!(fs::create_dir_all(&install_dir));
```
kpp: Do we really need `unwrap` crate?

libsodium-sys/build.rs
kpp: So you have to adopt their script https://github.com/maidsafe/rust_sodium/blob/master/rust_sodium-sys/src/lib.rs#L32 and import bindinds.rs into the repo. No need to regen them each build.

libsodium-sys/build.rs
neonphog: not my code... hesitant to change without understanding implications

kpp: Read https://crates.io/crates/unwrap . It's the same as .unwrap() but with additional info

libsodium-sys/build.rs
neonphog: woop, good catch, will fix

libsodium-sys/build.rs
neonphog: feels like this should be a separate effort, unless I'm missing something

libsodium-sys/build.rs
neonphog: mmm, Yeah, I see that now, will update

libsodium-sys/build.rs
neonphog: :+1: will do

libsodium-sys/build.rs
kpp: Yes. We can do that if you don't have time.

libsodium-sys/build.rs
```diff
@@ -63,34 +94,417 @@ fn find_libsodium_env() -> String {
...
     let statik = env::var_os("SODIUM_STATIC").is_some();
-    pkg_config::Config::new().statik(statik).probe("libsodium")
-        .expect("Could not find libsodium on this system");
-    pkg_config::get_variable("libsodium", "includedir")
-        .expect("Could not get includedir using pkg-config")
+    if let Err(e) = pkg_config::Config::new().statik(statik).probe("libsodium") {
+        println!("cargo:warning={:?}", e);
+        return None;
+    }
+    pkg_config::get_variable("libsodium", "includedir").ok()
```
kpp: Still you should warn "Could not get includedir using pkg-config" and return None

libsodium-sys/build.rs
```diff
@@ -63,34 +94,417 @@ fn find_libsodium_env() -> String {
...
 #[cfg(not(target_env = "msvc"))]
-fn find_libsodium_pkg() -> String {
+fn find_libsodium_pkg() -> Option<String> {
     let statik = env::var_os("SODIUM_STATIC").is_some();
-    pkg_config::Config::new().statik(statik).probe("libsodium")
-        .expect("Could not find libsodium on this system");
-    pkg_config::get_variable("libsodium", "includedir")
-        .expect("Could not get includedir using pkg-config")
+    if let Err(e) = pkg_config::Config::new().statik(statik).probe("libsodium") {
+        println!("cargo:warning={:?}", e);
```
kpp:
```
$ cargo build
   Compiling libsodium-sys v0.1.0 (/home/humbug/sodiumoxide/libsodium-sys)
warning: Failure { command: "\"pkg-config\" \"--libs\" \"--cflags\" \"libsodium\"", output: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "Package libsodium was not found in the pkg-config search path.\nPerhaps you should add the directory containing `libsodium.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'libsodium\' found\n" } }
   Compiling sodiumoxide v0.1.0 (/home/humbug/sodiumoxide)
```

You should add 1 more warning: "Trying to build from source"

libsodium-sys/build.rs
```diff
@@ -63,34 +94,417 @@ fn find_libsodium_env() -> String {
...
+    };
+    let include_dir = match lib.include_paths.get(0) {
+        Some(dir) => dir,
+        None => {
+            println!("cargo:warning=no include paths from vcpkg");
+            return None;
+        }
+    };
     include_dir.clone().into_os_string().into_string().ok()
-        .expect("Could not get cast include_dir to String")
```
kpp: You should keep this warning

libsodium-sys/build.rs
```diff
@@ -63,34 +94,417 @@ fn find_libsodium_env() -> String {
...
 Return: the first include path from vcpkg
 */
 #[cfg(target_env = "msvc")]
-fn find_libsodium_pkg() -> String {
-    let lib = vcpkg::probe_package("libsodium")
-        .expect("Could not find libsodium on this system");
-    let include_dir = lib.include_paths.get(0)
-        .expect("Could not get includedir using vcpkg");
+fn find_libsodium_pkg() -> Option<String> {
+    let lib = match vcpkg::probe_package("libsodium") {
```
kpp: https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else

```
let lib = vcpkg::probe_package("libsodium").unwrap_or_else(|e| {
    println!("cargo:warning={:?}", e);
    None
}).ok()?;

Comment (kpp) [inline libsodium-sys/build.rs]:
```
    let lib = vcpkg::probe_package("libsodium").map_err(|e| {
        println!("cargo:warning={:?}", e);
        e
    }).ok()?;
```

Maybe there is a more idiomatic way?

Comment (Dylan-DPC) [reply libsodium-sys/build.rs]: I think this is fine for a build script

Comment (kpp) [reply libsodium-sys/build.rs]: OK

File: libsodium-sys/build.rs (hunk 11 above)
Comment (neonphog) [reply libsodium-sys/build.rs]: added

File: libsodium-sys/build.rs (hunk 10 above)
Comment (neonphog) [reply libsodium-sys/build.rs]: added

File: libsodium-sys/build.rs (hunk 9 above)
Comment (neonphog) [reply libsodium-sys/build.rs]: added

File: libsodium-sys/build.rs (hunk 6 above)
Comment (kpp) [reply libsodium-sys/build.rs]: That's the whole point. See https://github.com/sodiumoxide/sodiumoxide/issues/280

Comment (kurnevsky) [reply libsodium-sys/build.rs]: Ok, but that means that static linking with downloaded libsodium should be default behavior. And perhaps build with different versions shouldn't be allowed. Otherwise everyone will use system libsodium ignoring it's version. I'm fine with statically generated bindings if we make these changes.

File: libsodium-sys/build.rs [hunk 13]
```diff
@@ -63,34 +97,430 @@ fn find_libsodium_env() -> String {
     } else {
         println!("cargo:rustc-link-lib={0}=sodium", mode);
     }
+}
+
+/* Must be called when SODIUM_LIB_DIR or SODIUM_INC_DIR is set to any value
+This function will set `cargo` flags.
+Return: SODIUM_INC_DIR
+*/
+fn find_libsodium_env() -> Option<String> {
```
kpp: A minor note: the function should return String as `build_libsodium` does.

Dylan-DPC:
bors: r+

Issue #256, opened 2018-09-26
Make --no-default-features great again


dnaq:
Either we need to fix so that sodiumoxide actually builds with --no-default-features (no-std) or we need to remove support for --no-default-features completely, which will clean up the codebase as well.

dnaq:
This should be easily solvable by removing our dependency on quickcheck, and all quickcheck tests completely. Those tests don't add anything to the library as is anyway.

dnaq:
Scratch that, we just need to import prelude::* before the tests when building with no-std

PR #283, opened 2018-11-22
Removed unnecessary dev-dependency on quickcheck

dnaq:
This dependency was only used for the tests for `verify32` and
`verify64` and the tests didn't test anything that wasn't already
tested.


kpp:
@dnaq does it close https://github.com/sodiumoxide/sodiumoxide/issues/256 ?

PR #300, opened 2018-12-27
WIP libsodium submodule

kpp:
Don't download libsodium artifacts. Build it from source!

TODO: sometimes `cargo build` fails if you rebuild the project. Repeat `cargo build` and it will be build. Weird.

Profit:
```
~/sodiumoxide$ cargo tree
sodiumoxide v0.2.0 (/home/humbug/sodiumoxide)                                                                                                        
├── libc v0.2.45
├── libsodium-sys v0.2.0 (/home/humbug/sodiumoxide/libsodium-sys)
│   └── libc v0.2.45 (*)
│   [build-dependencies]
│   ├── cc v1.0.28
│   └── pkg-config v0.3.14
└── serde v1.0.82
[dev-dependencies]
├── rmp-serde v0.13.7
│   ├── byteorder v1.2.7
│   ├── rmp v0.8.7
│   │   ├── byteorder v1.2.7 (*)
│   │   └── num-traits v0.1.43
│   │       └── num-traits v0.2.6
│   └── serde v1.0.82 (*)
├── rustc-serialize v0.3.24
├── serde v1.0.82 (*)
└── serde_json v1.0.33
    ├── itoa v0.4.3
    ├── ryu v0.2.7
    └── serde v1.0.82 (*)
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "libsodium-sys/libsodium"]
 	path = libsodium-sys/libsodium
-	url = git@github.com:jedisct1/libsodium.git
-	tag = 1.0.16
+	url = https://github.com/jedisct1/libsodium.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "libsodium-sys/libsodium"]
+	path = libsodium-sys/libsodium
+	url = git@github.com:jedisct1/libsodium.git
+	tag = 1.0.16
--- a/libsodium-sys/build.rs
+++ b/libsodium-sys/build.rs
@@ -1,37 +1,18 @@
-#[cfg(not(windows))]
 extern crate cc;
+
 #[cfg(not(target_env = "msvc"))]
-extern crate flate2;
-extern crate http_req;
-#[cfg(target_env = "msvc")]
-extern crate libc;
 extern crate pkg_config;
-extern crate sha2;
-#[cfg(not(target_env = "msvc"))]
-extern crate tar;
+
 #[cfg(target_env = "msvc")]
 extern crate vcpkg;
-#[cfg(target_env = "msvc")]
-extern crate zip;
 
-use http_req::request;
-use sha2::{Digest, Sha256};
 use std::env;
 use std::fs;
-use std::io::Cursor;
-use std::path::Path;
+use std::str;
+use std::path::{Path, PathBuf};
+use std::process::Command;
 
-static DOWNLOAD_BASE_URL: &'static str = "https://download.libsodium.org/libsodium/releases/";
 static VERSION: &'static str = "1.0.16";
-
-#[cfg(target_env = "msvc")] // libsodium-<VERSION>-msvc.zip
-static SHA256: &'static str = "0580d54f57594a7cb493607cec6e7045369fb67d43623491523781e901589948";
-
-#[cfg(all(windows, not(target_env = "msvc")))] // libsodium-<VERSION>-mingw.tar.gz
-static SHA256: &'static str = "5b81a4fc5d0de36dbda7efeaf355c133d4f6cc0b4dbf69bbe46ef7f5a6baa639";
-
-#[cfg(not(windows))] // libsodium-<VERSION>.tar.gz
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libsodium-sys/libsodium"]
+	path = libsodium-sys/libsodium
+	url = https://github.com/jedisct1/libsodium.git
```
kurnevsky: Shouldn't it be pinned to 1.0.16 tag?

kpp: There is no such mechanism to pin by tag, only by commit. See:

```
 Submodule libsodium added at 675149 
```

Which equals to 1.0.16

kurnevsky: ok, it seems you are right.

pcrockett:
So, being inexperienced in the world of C / C++, I'm noticing that I can't get this to build on my Ubuntu machine. My laptop is unable to run `autogen.sh`, and I don't know just yet what dependencies I'm missing or where to get them. I see no docs and I haven't fully researched it.

That said, if I download the released source directly and build it myself just using `./configure && make && make check`, my laptop builds the library just fine.

So my question is, with this change, how many extra dependencies would need to be installed on a Rust developer's machine in order to use sodiumoxide in their project?

Would it be possible to get the release source code in the `libsodium` directory, rather than using the libsodium Git repository?

jedisct1:
You should never run `autogen.sh`, unless you are making changes to the `configure.ac` code.

All the required files are already present both in releases and in the `stable` branch if you directly check out the git repository.

pcrockett:
> You should never run `autogen.sh`, unless you are making changes to the `configure.ac` code.
> 
> All the required files are already present both in releases and in the `stable` branch if you directly check out the git repository.

So then it looks like the submodule needs to be at commit 5bc564c, where 1.0.16 was merged into the stable branch. On my machine, I updated the submodule to that commit and removed the `autogen.sh` section from `build.rs`, and the build runs just fine.

kpp:
Well, I am not quite sure whether we should build it from sources.

@jedisct1 is there a universal script to build linsodium on any OS?

jedisct1:
@kpp There isn't, but `./configure && make check && make install` will usually do the job.

pcrockett:
@kpp Unless I'm mistaken, there's no way around building it from source for Unix-like OSes. Though building from source in Windows sounds like a pain. That looks like it requires installing the Windows 8.1 SDK. Would it be worth downloading the pre-compiled lib file when on Windows, but instead of using a SHA hash to verify it, use [the sequoia_openpgp crate][1] to verify the PGP signature?

[1]: https://docs.sequoia-pgp.org/sequoia_openpgp/

jedisct1:
@pcrockett @kpp Instead of relying on external tools, the signature can be verified with the [minisign_verify](https://docs.rs/minisign-verify) crate.

pcrockett:
@jedisct1 Of course... yeah the sequoia route is way overkill (and very beta). It didn't take me long to get minisign verification working based on the master branch. Decided to go ahead and open PR #329.

libsodium-sys/build.rs
```diff
@@ -142,186 +123,44 @@ fn find_libsodium_pkg() {
...
-            let include_file = entry_path.strip_prefix(arch_path).unwrap();
-            Path::new(&install_dir).join(include_file)
-        } else if entry_path == unpacked_lib {
-            lib_install_dir.join("libsodium.a")
-        } else {
-            continue;
-        };
-        entry.unpack(full_install_path).unwrap();
+    let source_dir = get_cur_dir().join("libsodium/");
+    if !source_dir.join(".git").exists() {
```
dnaq: I believe we shouldn't call out to git here. If someone builds our crate from crates.io the source should be bundled (maybe there's something we need to add to the manifest to make it so) and if someone clones the repo the readme should tell them to clone with --recursive.

libsodium-sys/build.rs
```diff
@@ -415,9 +254,27 @@ fn build_libsodium() {
         };
     }
 
+    // Run ./autogen.sh
```
dnaq: If we always pin to versions from the stable branch we should never need to run autogen, which is a lot better since there are systems that don't have autoconf/automake/libtool installed.

dnaq:
How do we handle builds for windows or other platforms. Also I would still like to keep the option of linking both dynamically and using a system libsodium.

PR #301, opened 2018-12-28
Add support for secretstream

### LINKED ISSUE #201: Expose bindings to crypto_secretstream_*()

dvic:
I have applied the generic push/pull Stream of @samscott89 (found in https://github.com/samscott89/sodiumoxide/tree/secretstream-take3) to the original branch of @nvesely (closed PR #240). I hope this is a good starting point for getting the secretstream API finally merged so we can close #201.


Dylan-DPC:
Thanks.
Waiting for reviews from @samscott89 and @dnaq in particular

dvic:
Does anyone know a better way to deal with `PhantomData` on `no-std`? My workaround feels a bit clunky.

samscott89:
> 
> 
> Does anyone know a better way to deal with `PhantomData` on `no-std`? My workaround feels a bit clunky.

It's in core :) - https://doc.rust-lang.org/core/marker/struct.PhantomData.html

src/crypto/secretstream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,242 @@
...
+        assert!(stream.rekey().is_err());
+    }
+
+    #[test]
+    fn tag_from_u8() {
+        assert_eq!(Tag::Message, Tag::from_u8(0).unwrap());
+        assert_eq!(Tag::Push, Tag::from_u8(1).unwrap());
+        assert_eq!(Tag::Rekey, Tag::from_u8(2).unwrap());
+        assert_eq!(Tag::Final, Tag::from_u8(3).unwrap());
+        for i in 4..(u16::from(std::u8::MAX) + 1) {
```
samscott89:
```suggestion
        for i in 4..(u16::from(core::u8::MAX) + 1) {
```

samscott89:
Many thank @dvic for picking this up and putting the PR together.

Dylan-DPC:
@dvic well yeah anyone from the team can review it, but since dnaq is the original author of this crate i prefer to rely on their reviews for such changes. Anyway, merging this

dnaq:
Sorry. I’ve been swamped.

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+/// `gen_key()` randomly generates a secret key
+///
+/// THREAD SAFETY: `gen_key()` is thread-safe provided that you have
+/// called `sodiumoxide::init()` once before using any other function
+/// from sodiumoxide.
+pub fn gen_key() -> Key {
+    let mut key = [0; KEYBYTES];
+    randombytes_into(&mut key);
```
kpp: Does it really have to be randombytes?

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -0,0 +1,353 @@
...
+        }
+        Ok((buf, tag))
+    }
+}
+
+// As additional precaution, rotate the keys when dropping the `Stream`
+// to ensure keys do no stay in memory.
+impl<T: StreamMode> Drop for Stream<T> {
+    fn drop(&mut self) {
+        let _ = self.rekey();
```
kpp: Does `new_type! secret Key(KEYBYTES);` add impl Drop for Key with drop=sodium_memzero?

src/crypto/secretstream/secretstream_macros.rs
dvic: What do you mean exactly? You mean to use a different method? I used here what was used in other modules to generate a key.

src/crypto/secretstream/secretstream_macros.rs
dvic: Yes (see [here](https://github.com/sodiumoxide/sodiumoxide/blob/7512559dc5c5404dc8e3edbda76e1297a393dc11/src/newtype_macros.rs#L213)).

kpp: So there is no sense in `let _ = self.rekey();` with comment `// to ensure keys do no stay in memory.`

src/crypto/secretstream/secretstream_macros.rs
kpp: OK.

dvic: AFAIK that is totally fine (it's what's also done in libsodium, see [here](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_secretstream/xchacha20poly1305/secretstream_xchacha20poly1305.c#L36). The [docs](https://libsodium.gitbook.io/doc/secret-key_cryptography/secretstream#encryption) also explain that any key can be used (the function referred to is `crypto_secretstream_xchacha20poly1305_keygen`)

> Note that using this function is not required to obtain a suitable key: the secretstream API can use any secret key whose size is crypto_secretstream_xchacha20poly1305_KEYBYTES bytes.
> Network protocols can leverage the key exchange API in order to get a shared key that can be used to encrypt streams. Similarly, file encryption applications can use the password hashing API to get a key that can be used with the functions below.

src/crypto/secretstream/secretstream_macros.rs
dvic: I guess the reason why this is here is the make sure the current internal key (stored in `crypto_secretstream_xchacha20poly1305_state`) is replaced by a 'useless' key. The new type `Key` is not used here (`rekey` returns `Result<(), ()>`).

src/crypto/secretstream/secretstream_macros.rs
kpp:
```
void
crypto_secretstream_xchacha20poly1305_keygen
   (unsigned char k[crypto_secretstream_xchacha20poly1305_KEYBYTES])
{
    randombytes_buf(k, crypto_secretstream_xchacha20poly1305_KEYBYTES);
}
```

So there is a special function to generate a key even though it equals to randombytes_into. Assume there will be another module and there will be another `crypto_secretstream_MODULE_keygen` with a different strategy. Is it possible?

dvic: Yes, I believe every module has its own `keygen()`, one example of this is in the `auth` module: https://github.com/sodiumoxide/sodiumoxide/blob/master/src/crypto/auth/auth_macros.rs#L36

kpp: If you use stream_module! on a module with keygen!=randombytes, the logic will be incorrect. If there is a stream module with such property, we should add crypto_secretstream_xchacha20poly1305_keygen as a $variable in macros. If there is no such a module, we shouldn't care.

dvic: As far as I know, all `keygen()` functions are always random bytes (except for keypair's of course). I guess they are redefined in every module for consistency reasons and the fact that the key sizes can differ.

dvic:
> @dvic well yeah anyone from the team can review it, but since dnaq is the original author of this crate i prefer to rely on their reviews for such changes. Anyway, merging this

@Dylan-DPC My guess is you haven't finished the sentence 😉? @dnaq will you have time this week or the next to look at this PR?

Dylan-DPC:
@dvic haha nope. "merging this" was all i wanted to write. I am waiting for the other co-maintainers to finish with their review and approve the PR and will merge it after that.

src/crypto/secretstream/secretstream_macros.rs
kpp: Ok, resolved

src/crypto/secretstream/secretstream_macros.rs
dvic: Ah I see, I've now removed the redundant rekeying in 2f0500e2c45e26122ba2f01153a64968073a9a89.

Dylan-DPC:
bors: r+

PR #303, opened 2019-01-02
Fix alignment of crypto_generichash_blake2b_state

dvic:
Without this fix, the added test produces a `SIGSEGV` on my machine. It's the same test as `test_vector_1` but with a `map_err` on the result of `finalize`.

I don't really understand why this is causing a `SIGSEGV` in the first place and I also don't understand why unpacking the struct in `finalize` fixes the problem. Can anyone confirm/explain the issue here?


dnaq:
It’s probably an alignment issue. The fix just happens to work since the
compiler happens to align the struct differently.

The princess is in another castle.

dvic:
A bit more context:
```
Process 24992 stopped
* thread #2, name = 'crypto::generichash::test::test_vector_1', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x00000001001dac66 sodiumoxide-2ad48efd1c6a7caa`blake2b_compress_ref + 166
sodiumoxide-2ad48efd1c6a7caa`blake2b_compress_ref:
->  0x1001dac66 <+166>: movdqa 0x40(%rbx), %xmm1
    0x1001dac6b <+171>: movdqa 0x50(%rbx), %xmm3
    0x1001dac70 <+176>: pxor   0x9c428(%rip), %xmm3      ; ge25519_select_base.base + 30734
    0x1001dac78 <+184>: pextrq $0x1, %xmm3, -0xb8(%rbp)
Target 0: (sodiumoxide-2ad48efd1c6a7caa) stopped.
```

With `cargo test --release` everything is fine, also running the tests using the docker `rust` image work fine. Is this a bug in Rust itself? Should I open an issue somewhere?

dvic:
@dnaq Thanks for the explanation! This was indeed the problem. I have pushed a patch and a test to assert the alignment of `crypto_generichash_state`.

I tried a new version of bindgen but it is not choosing the right alignment. My guess is that this is because of the preprocessor directive `CRYPTO_ALIGN` used in the definition of [crypto_generichash_state](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/include/sodium/crypto_generichash_blake2b.h#L23) (but I'm not sure how bindgen computes the alignment).

kpp:
Did it help?

dvic:
Yes, it solved the SIGSEGV on my machine (without any additional changes).

libsodium-sys/src/gen.sh
```diff
@@ -9,3 +9,5 @@ bindgen PATH_TO/libsodium-1.0.16/src/libsodium/include/sodium.h -o sodium_bindin
   --whitelist-function=$REGEX \
   --whitelist-type=$REGEX \
   --whitelist-var=$REGEX
+
+git apply crypto_generichash_blake2b_state.patch
```
kpp: Add new line

src/crypto/generichash/mod.rs
```diff
@@ -108,6 +108,11 @@ mod test {
     #[cfg(not(feature = "std"))]
     use prelude::*;
 
+    #[test]
+    fn test_crypto_generichash_state_alignment() {
+        assert_eq!(64, mem::align_of::<crypto_generichash_state>());
```
kpp: Add comment why it should be 64 aligned

dvic: Done, let me know if you have suggestions for anything you would like add or change.

src/crypto/generichash/mod.rs
```diff
@@ -108,6 +108,16 @@ mod test {
     #[cfg(not(feature = "std"))]
     use prelude::*;
 
+    #[test]
+    fn test_crypto_generichash_state_alignment() {
+        // bindgen does not properly compute the alignment of crypto_generichash_state,
+        // which is why gen.sh in libsodium-sys applies a patch after it has run rust-bindgen
```
kpp: That should be a comment for gen.sh =)

The comment inside this test should correspond to libsodium docs where it is required for crypto_generichash_blake2b_state to be 64-aligned with the link to libsodium's docs

dvic:
Closing in favor of #304

kpp:
What's the reason for closing this PR?

PR #312, opened 2019-02-15
Reduce deps to build on linux from 61 -> to 48

kpp:
When call `cargo build` the total number of crates to build is 48 (was 61).
When call `cargo tree | wc -l` the result is 74 (was 94).

### DIFF
--- a/libsodium-sys/Cargo.toml
+++ b/libsodium-sys/Cargo.toml
@@ -15,20 +15,24 @@ version = "0.2.0"
 travis-ci = { repository = "sodiumoxide/sodiumoxide" }
 
 [build-dependencies]
-cc = "1.0.9"
-flate2 = "1.0.1"
-http_req = "0.2.1"
-pkg-config = "0.3.11"
-libc = { version = "0.2.41" , default-features = false }
+http_req = { version = "0.4", default-features = false, features = ["rust-tls"] }
 sha2 = "0.8"
-tar = "0.4.15"
-zip = "0.5"
+pkg-config = "0.3"
+
+
 
 [target.'cfg(target_env = "msvc")'.build-dependencies]
+libc = { version = "=0.2.48" , default-features = false }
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
while we are it, can we avoid depending on the entire version? rust will consider it as `^x.y.z ` but i think it is better to go with `x.y` or `x` for 1.0+ crates so that we are more flexible

kpp:
> to go with `x.y` or `x` for 1.0+ crates so that we are more flexible

Done

Dylan-DPC:
i think we need to make an exception for libc and depend on the full version (sorry :sweat) because it often breaks so it is better if everyone has the same version

kpp:
> depend on the full version

Done

Dylan-DPC:
bors: r+

libsodium-sys/Cargo.toml
```diff
@@ -15,20 +15,24 @@ version = "0.2.0"
...
+zip = "0.5"
+
+[target.'cfg(not(target_env = "msvc"))'.build-dependencies]
+cc = "1.0"
+libflate = "0.1"
+tar = "0.4"
 
 [dependencies]
-libc = { version = "^0.2.41" , default-features = false }
+libc = { version = "=0.2.48" , default-features = false }
```
kpcyrd: Is there a reason libc was pinned? This prevents libc updates for all projects depending on sodiumoxide.

kpp: Why? IIRC cargo allows you to have multiple versions of the same library in you tree

kpcyrd: I think there's something special with `0.2.X` versions, adding and removing sodiumoxide as a dependency causes the libc dependency to flip back and forth between 0.2.48 and 0.2.49 for all crates. It also breaks as soon as any dependency in the project depends on `libc = "0.2.49"`.

PR #315, opened 2019-02-25
Add use-pkg-config feature

kpcyrd:
This feature allows a crate to depend on sodiumoxide and force dynamic linking. This allows a project to revert to the old default without using a non-standard rust build system.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,3 +38,4 @@ rmp-serde = "^0.13.7"
 benchmarks = []
 std = []
 default = ["serde", "std"]
+use-pkg-config = ["libsodium-sys/use-pkg-config"]
--- a/libsodium-sys/Cargo.toml
+++ b/libsodium-sys/Cargo.toml
@@ -36,3 +36,6 @@ libc = { version = "=0.2.48" , default-features = false }
 
 [lib]
 name = "libsodium_sys"
+
+[features]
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -38,3 +38,4 @@ rmp-serde = "^0.13.7"
 benchmarks = []
 std = []
 default = ["serde", "std"]
```
kpp: How about adding use-pkg-config to defaults?

kpcyrd: That would be my preferred solution, this would be a breaking change though. Do we want to do that regardless? Do we still want to allow control with an environment variable like `SODIUM_USE_PKG_CONFIG`?

Dylan-DPC: I don't think it is something we need in default features at the moment. It is fine as is.

Dylan-DPC:
@kpp can you re-review this?

Cargo.toml
kpp: Ah, I see. If feature `use-pkg-config` isset we should use pkg config, otherwise fallback to the previous solution with SODIUM_USE_PKG_CONFIG. Resolved

Dylan-DPC:
bors: r+

Dylan-DPC:
bors: r+

PR #317, opened 2019-02-27
Unpin libc dependency

kpcyrd:
This makes sure we don't run into unresolvable dependency conflicts if any crate depends on `libc = "> 0.2.48"`.

### DIFF
--- a/libsodium-sys/Cargo.toml
+++ b/libsodium-sys/Cargo.toml
@@ -22,7 +22,7 @@ pkg-config = "0.3"
 
 
 [target.'cfg(target_env = "msvc")'.build-dependencies]
-libc = { version = "=0.2.48" , default-features = false }
+libc = { version = "0.2" , default-features = false }
 vcpkg = "0.2"
 zip = "0.5"
 
@@ -32,7 +32,7 @@ libflate = "0.1"
 tar = "0.4"
 
 [dependencies]
-libc = { version = "=0.2.48" , default-features = false }
+libc = { version = "0.2" , default-features = false }
 
 [lib]
 name = "libsodium_sys"


kpp:
@kpcyrd would you please point why is it important to merge

Dylan-DPC:
Will explain the use case / problem of this when I get to office. Thanks

dnaq:
I would also like to see this merged before we cook the release.

kpp:
@dnaq why?

dnaq:
To avoid users of the new release linking in multiple versions of libc when
using sodiumoxide. And also to avoid the issues with unresolvable
dependency conflicts. As soon as the library is on crates.io it will be the
version that people use, so dependency conflicts will generate bug reports.

Dylan-DPC:
Yes that's right we need to avoid users having conflicts with libc. But at the same time, libc often breaks in point releases, so it also makes sense to pin to a particular version. I will speak to the co-maintainers of libc and get back on this later today or tomorrow.

Dylan-DPC:
Okay, i'm fine with this change.

Dylan-DPC:
Had a discussion with libc maintainers, they are moving towards making libc semver-compatible. So we are good. in case there are any breaking changes, we need to issue patch releases that pin in to a version where we are fine with libc and report it upstream.

libsodium-sys/Cargo.toml
```diff
@@ -32,7 +32,7 @@ libflate = "0.1"
 tar = "0.4"
 
 [dependencies]
-libc = { version = "=0.2.48" , default-features = false }
+libc = { version = "0.2.48" , default-features = false }
```
Dylan-DPC: we don't need to pin to this version. we can go with "0.2"

kpp: Why? Let it be

kpcyrd: This doesn't pin it, but defines a minimum version. Should I change it regardless?

Dylan-DPC: this could still cause conflicts or have multiple version if a crate uses libc < 0.2.48. So unless there is a good reason to pin to 0.2.48 (such as a feature that was introduced in that release or a fix) it is better to be more "liberal" by specifying 0.2

Dylan-DPC: ouch we were both typing at the same time :P well yeah not a "pin" but a lower limit, but it just reduces the versions that could collide. I'd prefer 0.2 but either way i'm good with it

Dylan-DPC:
bors: r+

kpp:
bors: r+

PR #323, opened 2019-03-28
Add `AsRef<[u8]>` for newtypes

### LINKED ISSUE #321: Expose `AsRef<[u8]>` for all newtypes
Hi,

According to the [libsodium documentation](https://docs.rs/sodiumoxide/0.2.1/sodiumoxide/crypto/kx/index.html#usage):

> These keys can be used by any functions requiring secret keys up to crypto_kx_SESSIONKEYBYTES bytes, including crypto_secretbox_*() and crypto_aead_*().

Therefore at least the AsRef trait must be implemented for the SessionKey type to be able to convert it to the `secretbox::Key` (or `secretstream::Key` in my use case). Otherwise it looks like once SessionKey is generated, there's nothing that can be done with it. Please correct me if I'm wrong here.

As far as I can tell this is because implementation of SessionKey is generated via a "secret" key macro.  Should it be "public" instead?

Cheers,
Mike

dnaq:
This seems like a much cleaner way to access the raw byte contents of newtypes instead
of exposing `Index` implementations.

We also deprecate the `Index` implementations (even though deprecation on trait-impls doesn't do anything for now). We should probably remove these implementations in the next major release.

Closes #321.

### DIFF
--- a/src/newtype_macros.rs
+++ b/src/newtype_macros.rs
@@ -82,13 +82,6 @@ macro_rules! newtype_traits (($newtype:ident, $len:expr) => (
         }
     }
 
-    impl AsMut<[u8]> for $newtype {
-        #[inline]
-        fn as_mut(&mut self) -> &mut [u8] {
-            &mut self.0
-        }
-    }
-
     /// Allows a user to access the byte contents of an object as a slice.
     ///
     /// WARNING: it might be tempting to do comparisons on objects
--- a/src/newtype_macros.rs
+++ b/src/newtype_macros.rs
@@ -39,7 +39,7 @@ macro_rules! newtype_traits (($newtype:ident, $len:expr) => (
         fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
             where S: ::serde::Serializer
         {
-            serializer.serialize_bytes(&self[..])
+            serializer.serialize_bytes(&self.as_ref())
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/newtype_macros.rs
```diff
@@ -75,12 +75,27 @@ macro_rules! newtype_traits (($newtype:ident, $len:expr) => (
...
+    impl AsRef<[u8]> for $newtype {
+        #[inline]
+        fn as_ref(&self) -> &[u8] {
+            &self.0
+        }
+    }
+
+    impl AsMut<[u8]> for $newtype {
+        #[inline]
+        fn as_mut(&mut self) -> &mut [u8] {
```
kpp: Does this impl provides a write access to a type that was previously impossible to do? Can I change the inner data of a newtype without this PR?

dnaq: You could do it before, using the index-implementation, or you could create a new one using `from_slice()`. I'm ok with that, given that people might want to read keys from files or other sources.

kpp: You have to use `std::ops::IndexMut` to write into the struct. Was there such an impl before your PR?

dnaq: No, you’re right. There wasn’t.

kpp: I think there should not be AsMut impl for newtype. If you want to change the struct, use ::from_slice and replace the whole struct.

dnaq:
I’m ok with removing AsMut for now

PR #324, opened 2019-03-29
Implement memory locking and unlocking

oherrala:
This PR implements two utils: `mlock()` and `munlock()` which are direct wrappers for `sodium_mlock()` and `sodium_munlock()`.

The unit test is pretty silly, but I'm not sure what we could test for `mlock`.

libsodium recommends using these functions for password (https://libsodium.gitbook.io/doc/advanced/scrypt#notes):

> It is highly recommended to use sodium_mlock() to lock memory regions storing cleartext passwords, and to call sodium_munlock() right after crypto_pwhash_scryptsalsa208sha256_str() and crypto_pwhash_scryptsalsa208sha256_str_verify() return.
>
> sodium_munlock() overwrites the region with zeros before unlocking it, so it doesn't have to be done before calling this function.

My use case is to use these functions for implementing `struct Password(Vec<u8>)` with appropriate locking and `Drop` trait implementation doing zeroing and unlocking of `Password`.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -25,6 +25,40 @@ pub fn memcmp(x: &[u8], y: &[u8]) -> bool {
     unsafe { ffi::sodium_memcmp(x.as_ptr() as *const _, y.as_ptr() as *const _, x.len()) == 0 }
 }
 
+
+/// `mlock()` locks memory given region which can help avoiding swapping the
+/// sensitive memory region to disk.
+///
+/// Operating system might limit the amount of memory a process can `mlock()`.
+/// This function can fail if `mlock()` fails to lock the memory.
+pub fn mlock(x: &mut [u8]) -> Result<(), ()> {
+    let ret = unsafe {
+        ffi::sodium_mlock(x.as_mut_ptr() as *mut _, x.len())
+    };
+    if ret == 0 {
+        Ok(())
+    } else {
+        Err(())
+    }
+}
+
+/// `munlock()` unlocks memory region.
+///
+/// `munlock()` overwrites the region with zeros before unlocking it, so it
+/// doesn't have to be done before calling this function.
+pub fn munlock(x: &mut [u8]) -> Result<(), ()> {
+    let ret = unsafe {
+        // sodium_munlock() internally calls sodium_memzero() to clear memory
+        // region
+        ffi::sodium_munlock(x.as_mut_ptr() as *mut _, x.len())
+    };
+    if ret == 0 {
+        Ok(())
+    } else {
+        Err(())
+    }
+}
+
 /// `increment_le()` treats `x` as an unsigned little-endian number and increments it in
 /// constant time.
 ///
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -202,4 +233,15 @@ mod test {
             assert_eq!(x, vec![1u8; i]);
         }
     }
+
+    #[test]
+    fn test_mlock_munlock() {
+        let t = b"hello world";
+        let mut x = Vec::new();
+        x.extend_from_slice(t);
+        assert!(mlock(&mut x).is_ok());
```
kpp: How about tests for `is_err`?

src/utils.rs
```diff
@@ -202,4 +233,15 @@ mod test {
...
+        assert_eq!(&x, t);
+        assert!(munlock(&mut x).is_ok());
```
kpp: How about tests for `is_err`?

src/utils.rs
oherrala: Both `sodium_mlock()` and `sodium_munlock()` are platform dependent functions whose failure is operating system and local configuration dependent. I expect most of the error conditions are not possible with rust allocated `&[u8]` and triggering the system or per-process limit for locked memory probably needs complicated test setup for setting limits.

If there's error conditions which we could trigger, I'm more than happy to try implementing those.

Ref:
 * https://github.com/jedisct1/libsodium/blob/a6ef940634c6191ceb36a73f8c355655d4d6c740/src/libsodium/sodium/utils.c#L406-L420
 * https://github.com/jedisct1/libsodium/blob/a6ef940634c6191ceb36a73f8c355655d4d6c740/src/libsodium/sodium/utils.c#L422-L437

kpp: How about mlocking it twice or trying to munlock a non-mlocked memory region?

oherrala: double locking or unlocking not-locked memory in macOS work fine. No errors there. Reading the POSIX IEEE Std 1003.1-2017 and man pages from Linux, OpenBSD and macOS seem to not document such error conditions.

Also from Linux man page:

> Memory locks do not stack, that is, pages which have been locked several times by calls to mlock(), mlock2(), or mlockall() will be unlocked by a single call to munlock() for the corresponding range or by munlockall().

Ref. http://man7.org/linux/man-pages/man2/mlock.2.html

kpp: Ah. How do we test it then?

oherrala: I added a unix specific test which uses [`setrlimit(2)`](https://linux.die.net/man/2/setrlimit) to set a limit for locked memory size which then causes `mlock()` to fail. I tested this with Linux and macOS. See https://github.com/sodiumoxide/sodiumoxide/pull/324/commits/68f5a64b32e48c1bae3bbc26fc568448d8ea56e9.

However, this feels a bit wrong since it's success depends on compile time parameters of libsodium itself. If libsodium gets compiled without mlock() support, then this test fails. See https://github.com/jedisct1/libsodium/blob/a6ef940634c6191ceb36a73f8c355655d4d6c740/src/libsodium/sodium/utils.c#L412-L413

Dylan-DPC:
bors: r+

PR #325, opened 2019-03-31
Initialize libsodium with global constructor

dnaq:
This frees the user of sodiumoxide from remembering to call `sodiumoxide::init()`, which we deprecate for now.

NOTE: maybe we shouldn't deprecate this function and instead keep it so that a caller can use it to make sure that initialization hasn't somehow failed? This should be discussed before merging this pull-request.

We could panic in the constructor instead, but that feels a bit scary since it runs pre-main, and we don't have any guarantees about the state of rust when that happend (I might be wrong here, please correct me if I am).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,6 +61,8 @@
 
 extern crate libsodium_sys as ffi;
 
+#[macro_use]
+extern crate ctor;
 extern crate libc;
 #[cfg(test)]
 extern crate rustc_serialize;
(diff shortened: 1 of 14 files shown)


kpp:
Nice!

src/lib.rs
```diff
@@ -97,6 +103,13 @@ pub fn init() -> Result<(), ()> {
     }
 }
 
+#[ctor]
+fn init_real() {
+    unsafe {
+        ffi::sodium_init();
```
kpp: @mmastrac can we fail here somehow?

mmastrac: As this is a pure FFI call, it's a safe way to use `#[ctor]`. If your users downstream try to use some of the bindings in their own `#[ctor]` code they will obviously need to be cautious as there's no guarantee to the initialization order (more of a doc issue/caveat).

Dylan-DPC:
bors: r+

kpp:
By the way, how about making it optional via features and disabling it by default to keep the original behavior?

PR #329, opened 2019-04-20
Verifying libsodium with minisign instead of sha2

### LINKED ISSUE #326: Failed hash check
I received this error when building on mingw64. I had to enable SODIUM_USE_PKG_CONFIG in order to successfully build. I checked the releases at https://download.libsodium.org/libsodium/releases/ and it looks like 1.0.17 was modified on 03-Apr-2019 so I think the checksum just needs to be updated. I don't have any experience in hash checking, so I might be wrong.

pcrockett:
This would resolve 326, and makes it so we don't have to keep updating hard-coded SHA hashes all the time. The build seems to run fine on Windows and Ubuntu.

Sorry if anything is out of order - I'm a bit new to Rust, not to mention contributing to open source projects in general.

### DIFF
--- a/libsodium-sys/Cargo.toml
+++ b/libsodium-sys/Cargo.toml
@@ -16,9 +16,8 @@ travis-ci = { repository = "sodiumoxide/sodiumoxide" }
 
 [build-dependencies]
 http_req = { version = "0.4", default-features = false, features = ["rust-tls"] }
-sha2 = "0.8"
 pkg-config = "0.3"
-
+minisign-verify = "0.1.1"
 
 
 [target.'cfg(target_env = "msvc")'.build-dependencies]
--- a/libsodium-sys/build.rs
+++ b/libsodium-sys/build.rs
@@ -5,8 +5,8 @@ extern crate http_req;
 extern crate libc;
 #[cfg(not(target_env = "msvc"))]
 extern crate libflate;
+extern crate minisign_verify;
 extern crate pkg_config;
-extern crate sha2;
 #[cfg(not(target_env = "msvc"))]
 extern crate tar;
 #[cfg(target_env = "msvc")]
@@ -15,7 +15,7 @@ extern crate vcpkg;
 extern crate zip;
 
 use http_req::request;
-use sha2::{Digest, Sha256};
+use minisign_verify::{PublicKey, Signature};
 use std::env;
 use std::fs;
 use std::io::Cursor;
@@ -23,15 +23,7 @@ use std::path::Path;
 
 static DOWNLOAD_BASE_URL: &'static str = "https://download.libsodium.org/libsodium/releases/";
 static VERSION: &'static str = "1.0.17";
-
-#[cfg(target_env = "msvc")] // libsodium-<VERSION>-msvc.zip
-static SHA256: &'static str = "66aa0ca60c171201a9e795b8abcd7881bc9f3c664f0f80a53a55c6723dd97ad7";
-
-#[cfg(all(windows, not(target_env = "msvc")))] // libsodium-<VERSION>-mingw.tar.gz
-static SHA256: &'static str = "ea385a26997f2d8f8ea44ed930da6fe233f9377753f0436e40f80f221ded190e";
-
-#[cfg(not(windows))] // libsodium-<VERSION>.tar.gz
-static SHA256: &'static str = "0cc3dae33e642cc187b5ceb467e0ad0e1b51dcba577de1190e9ffa17766ac2b1";
+static MINISIG_PUB_KEY: &'static str = "RWQf6LRCGA9i53mlYecO4IzT51TGPpvWucNSCh1CBM0QTaLn73Y7GFO3";
 
 fn main() {
     println!("cargo:rerun-if-env-changed=SODIUM_LIB_DIR");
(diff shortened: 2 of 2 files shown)


kpp:
Looks good. How does it affect total number of deps in dependency tree? See ‘cargo tree’

pcrockett:
The relevant part of the dependency tree looks like this:

```
├── minisign-verify v0.1.1
│   ├── base64 v0.10.1 (*)
│   └── ed25519-dalek v1.0.0-pre.1
│       ├── clear_on_drop v0.2.3
│       │   [build-dependencies]
│       │   └── cc v1.0.35 (*)
│       ├── curve25519-dalek v1.1.3
│       │   ├── byteorder v1.3.1 (*)
│       │   ├── clear_on_drop v0.2.3 (*)
│       │   ├── digest v0.8.0
│       │   │   └── generic-array v0.12.0
│       │   │       └── typenum v1.10.0
│       │   ├── rand_core v0.3.1
│       │   │   └── rand_core v0.4.0
│       │   └── subtle v2.0.0
│       │   [build-dependencies]
│       │   ├── byteorder v1.3.1 (*)
│       │   ├── clear_on_drop v0.2.3 (*)
│       │   ├── digest v0.8.0 (*)
│       │   ├── rand_core v0.3.1 (*)
│       │   └── subtle v2.0.0 (*)
│       ├── failure v0.1.5
│       ├── rand v0.6.5
│       │   ├── libc v0.2.51 (*)
│       │   ├── rand_chacha v0.1.1
│       │   │   └── rand_core v0.3.1 (*)
│       │   │   [build-dependencies]
│       │   │   └── autocfg v0.1.2
│       │   ├── rand_core v0.4.0 (*)
│       │   ├── rand_hc v0.1.0
│       │   │   └── rand_core v0.3.1 (*)
│       │   ├── rand_isaac v0.1.1
│       │   │   └── rand_core v0.3.1 (*)
│       │   ├── rand_jitter v0.1.3
│       │   │   └── rand_core v0.4.0 (*)
│       │   ├── rand_os v0.1.3
│       │   │   ├── libc v0.2.51 (*)
│       │   │   └── rand_core v0.4.0 (*)
│       │   ├── rand_pcg v0.1.2
│       │   │   └── rand_core v0.4.0 (*)
│       │   │   [build-dependencies]
│       │   │   └── autocfg v0.1.2 (*)
│       │   └── rand_xorshift v0.1.1
│       │       └── rand_core v0.3.1 (*)
│       │   [build-dependencies]
│       │   └── autocfg v0.1.2 (*)
│       └── sha2 v0.8.0
│           ├── block-buffer v0.7.3
│           │   ├── block-padding v0.1.3
│           │   │   └── byte-tools v0.3.1
│           │   ├── byte-tools v0.3.1 (*)
│           │   ├── byteorder v1.3.1 (*)
│           │   └── generic-array v0.12.0 (*)
│           ├── digest v0.8.0 (*)
│           ├── fake-simd v0.1.2
│           └── opaque-debug v0.2.2
│           [dev-dependencies]
│           └── digest v0.8.0 (*)
```
Notice that the sha2 package is still in there even though we're not depending on it directly. I count 20 new dependencies that we weren't already using (without an `(*)`).

kpp:
That’s a shame we need so many deps to verify hashes :(

dnaq:
The ironic part is that minisign would be easy to implement without further dependencies if we had access to libsodium :)

jedisct1:
These are just build dependencies, and are required by `ed25519-dalek`. I can get rid of `ed25519-dalek`, but is it worth it, especially for a build dependency?

kpp:
> but is it worth it, especially for a build dependency?

https://github.com/tox-rs/tox-node already got 195 deps and roughly half of them are ssl, http, tar and sha2 to download and verify libsodium. I thought it did not matter until I tried to compile tox-node under AWS micro and it took 12 minutes.

I believe we don't have an option, so it's a yes from me. A sad but yes.

kpcyrd:
Keep in mind that this resolves the build failure, but the root cause that we're using mutable data in our build makes the build nondeterministic from a reproducible builds point of view. To guarantee independent verification of the binary we would need to track which rebuild of the library release was used during our build or make sure all inputs of the build are immutable (eg. by building libsodium from immutable source releases).

More on that topic can be found here: https://reproducible-builds.org/

kpcyrd:
Also note that this allows a subtle way for the owner of the minisign private key to carry out very targeted attacks by serving a malicious pre-built library with a valid signature. Since there's no binary transparency you can't be sure you are getting the same library everybody else is getting unless we're pinning the file content by hash.

jedisct1:
Version `0.1.2` of the crate doesn't have any dependency any more besides `base64`.

kpp:
> serving a malicious pre-built library with a valid signature

That is so true.

pcrockett:
That is a really good point about reproducible builds. Especially when considering what kind of downstream projects could be using this library, reproducible builds would be a good feature to provide. I wonder if it would be worthwhile maintaining our own libsodium repo by forking it, and using the GitHub Releases feature to host libsodium releases in an immutable way (i.e. `1.0.17.1`, `1.0.17.2`, etc.).

And to be clear, I'm not suggesting forking the _project_. Just forking the _repository_ and keeping it in sync with the original.

kpp:
Closed in favor of https://github.com/sodiumoxide/sodiumoxide/pull/332

PR #331, opened 2019-04-27
Library bundle

kpp:


libsodium-sys/src/sodium_bindings.rs
```diff
@@ -1,5 +1,7 @@
 /* automatically generated by rust-bindgen */
 
+use libc;
```
kpp: Any changes made to this file will be overwritten when we regenerate bindings

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -128,7 +129,7 @@ pub fn gen_key() -> Key {
 pub struct Stream<M: StreamMode> {
     state: $state_name,
     finalized: bool,
-    phantom: core::marker::PhantomData<M>,
```
kpp: `core::marker` is not an old Rust it is the way to use sodiumoxide in a no-std env

kpp:
Closed in favor of https://github.com/sodiumoxide/sodiumoxide/pull/332

PR #337, opened 2019-05-12
Add clippy to CI

### LINKED ISSUE #236: clippy
`cargo clippy --all --tests --all-features -- -D clippy` should be run with a pinned clippy version.

We should use the `clippy_pedantic` setting via a `clippy.toml`. While it may occasionally throw out a false positive, I think it's better just to use an ignore attribute appropriately, than to sacrifice some delicious pedantry :japanese_ogre: 

See the reasoning in #235 for why clippy should be pinned. Since it's installed a different way than rustfmt, you want to pin the version of clippy and the nightly (the nightly doesn't determine the clippy version, but if you pin clippy, you also want to pin a nightly that you know works with it).

See https://github.com/cad97/great-ci/ and https://github.com/SpinResearch/RustySecrets/blob/master/.travis.yml for some examples of how to do this in Travis (also useful for #235) while making travis run as fast as possible. One night for cacheing cargo is that you should actually run `cargo update` in the build script if you're going to do that.

kpp:
Closes #236 

`#![deny(clippy::pedantic)]` is too agressive and we have to switch off a lot of lints from pedantic groups like:

```
error: item name starts with its containing module's name
  --> src/randombytes.rs:25:1
   |
25 | / pub fn randombytes_into(buf: &mut [u8]) {
26 | |     unsafe {
27 | |         ffi::randombytes_buf(buf.as_mut_ptr() as *mut _, buf.len());
28 | |     }
29 | | }
   | |_^
   |
note: lint level defined here
  --> src/lib.rs:62:9
   |
62 | #![deny(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: #[deny(clippy::module_name_repetitions)] implied by #[deny(clippy::pedantic)]
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions

error: aborting due to 204 previous errors
```

So I decided not to include pedantic lints. Because we will have to either switch pedantic lints off or fix 204 errors and break the API.

Unresolved issue: https://github.com/sodiumoxide/sodiumoxide/issues/338


src/crypto/generichash/mod.rs
```diff
@@ -155,7 +155,8 @@ mod test {
         loop {
             let msg = {
                 line.clear();
-                if let Err(_) = r.read_line(&mut line) {
+                r.read_line(&mut line).unwrap();
+                if line.is_empty() {
```
dnaq: Are these semantically equivalent?

src/crypto/sign/ed25519.rs
```diff
@@ -477,7 +477,7 @@ mod test {
         let mut creation_state = State::init();
         creation_state.update(&m);
 
-        let creation_state_cloned = creation_state.clone();
+        let creation_state_cloned = creation_state;
```
dnaq: This should have a different name

src/crypto/generichash/mod.rs
kpp: I took the same code from https://github.com/sodiumoxide/sodiumoxide/pull/337/files#diff-c75fa680a5ac2678ede6c3c657ef37b0L71 and https://github.com/sodiumoxide/sodiumoxide/pull/337/files#diff-e56ac0ce53049e9b4c3da2aaac1f530aR56 .

src/crypto/sign/ed25519.rs
kpp: Wow. The name of the test is `test_streaming_clone`. So it tests whether we can clone it. I will rename it to `test_streaming_copy` and `creation_state_cloned` -> `creation_state_copy`. How does it sound?

src/crypto/generichash/mod.rs
kpp: I took the same code from https://github.com/sodiumoxide/sodiumoxide/pull/337/files#diff-c75fa680a5ac2678ede6c3c657ef37b0L71 and https://github.com/sodiumoxide/sodiumoxide/pull/337/files#diff-e56ac0ce53049e9b4c3da2aaac1f530aR56 .

src/crypto/sign/ed25519.rs
kpp: Ah. There is already such a test `test_streaming_copy`.

src/crypto/generichash/mod.rs
kpp: Oh you are right. There are empty lines in testvectors/blake2b-kat.txt. I tested them, if fileline is empty, then line="\n", so line.is_empty() only at the EOF.

> Are these semantically equivalent?

yes

src/crypto/sign/ed25519.rs
kpp: I removed the test for clone()

kpp:
bors: r+

PR #342, opened 2019-06-05
Update libsodium to 1.0.18

kurnevsky:
Alignment patch is no longer needed with this update.


libsodium-sys/build.rs
```diff
@@ -146,12 +146,12 @@ fn build_libsodium() {
...
     if is_release_profile() {
         println!(
-            "cargo:rustc-link-search=native={}/msvc/Win32/Release/v140/",
+            "cargo:rustc-link-search=native={}/msvc/Win32/Release/v142/",
             get_crate_dir()
         );
     } else {
         println!(
-            "cargo:rustc-link-search=native={}/msvc/Win32/Debug/v140/",
+            "cargo:rustc-link-search=native={}/msvc/Win32/Debug/v142/",
```
kpp: Why 142?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Add `AsRef<[u8]>` for newtypes (#323)
 * Implement memory locking and unlocking (#324)
 * Add convenience functions for pwhash() and derive_key() (#309)
+* Update libsodium to 1.0.18 (#342)
```
kpp: 0.2.2 was released on May 16, 2019, so you have to create a new patch in changelog.

libsodium-sys/build.rs
kurnevsky: Isn't it a some kind of a version of a library? I just took the latest one which is 142.

kpp: https://habr.com/ru/company/microsoft/blog/439664/

kurnevsky: So that means that older msvc compiler won't be able to use this version? An we still want to use 140?

kpp: > We want to keep the momentum going and make sure that you have a similarly successful adoption experience with MSVC v142 too. This is why we’re announcing today that our team is committed to provide binary compatibility for MSVC v142 with both MSVC v141 and v140.
This means that if you compile all your code with the v142 toolset but still have one or more libraries that are built with the v140 or v141 toolset, linking all of it together (with the latest linker) will work as expected. To make this possible, VC Runtime does not change its major version in VS 2019 and remains backward compatible with previous VC Runtime versions.

There is a backward compatibility, but I don't see anything about forward compatibility.

kurnevsky: ok, reverted to 140.

CHANGELOG.md
kurnevsky: Fixed.

kurnevsky:
> What is the reason for it not being needed though?

I'm not sure but it doesn't really matter. We have committed bindings into the repo so they always will be correct once generated correctly. Because of this we support only one version of libsodium (but anyway alignments won't break even for different versions). Also we have manually written test for alignments so it's even not possible to generate incorrect bindings without being noticed: https://github.com/sodiumoxide/sodiumoxide/blob/69a69fb091c59cb853230cee5a1e4d58780ca5a8/libsodium-sys/tests/crypto/crypto_generichash_blake2b.rs#L8

dnaq:
That’s true. I’ll approve the pull request. However I would be even happier if we could add a test to check the alignment so we don’t get any regressions next time we regenerate the bindings.

kurnevsky:
But we have it, see the link from my previous comment :)

dnaq:
Sorry. I’m on mobile and a bit sleep deprived. :)

kurnevsky:
It seems it was bindgen: https://github.com/rust-lang/rust-bindgen/blob/master/CHANGELOG.md#0480

Fixed comments.

dnaq:
Then we should add an updated min dependency for bindgen

kurnevsky:
We don't depend on bindgen anymore and don't specify it's version anywhere - it's used only when bindings are generated.

kpp:
>  it's used only when bindings are generated.

Add a comment with min bindgen version to .sh file

Issue #338, opened 2019-05-12
UB in libsodium-sys/tests/crypto/crypto_generichash


kpp:
Run:

```
$ cargo clippy --all --tests
```

Which gives you a warning:

```
warning: transmute from a pointer to a pointer
  --> libsodium-sys/tests/crypto/crypto_generichash.rs:99:24
   |
99 |     let pst = unsafe { mem::transmute::<*mut u8, *mut crypto_generichash_state>(st.as_mut_ptr()) };
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `st.as_mut_ptr() as *mut libsodium_sys::crypto_generichash_blake2b_state`
   |
```

View in a Travis build: https://travis-ci.org/sodiumoxide/sodiumoxide/jobs/531384794#L502

Fixing it as suggested:

```diff
diff --git a/libsodium-sys/tests/crypto/crypto_generichash.rs b/libsodium-sys/tests/crypto/crypto_generichash.rs
index e40e3bae..d744c35f 100644
--- a/libsodium-sys/tests/crypto/crypto_generichash.rs
+++ b/libsodium-sys/tests/crypto/crypto_generichash.rs
@@ -96,7 +96,7 @@ fn test_crypto_generichash_multipart() {
     let key = [0u8; crypto_generichash_KEYBYTES as usize];
 
     let mut st = vec![0u8; unsafe { crypto_generichash_statebytes() }];
-    let pst = unsafe { mem::transmute::<*mut u8, *mut crypto_generichash_state>(st.as_mut_ptr()) };
+    let pst = st.as_mut_ptr() as *mut libsodium_sys::crypto_generichash_blake2b_state;
 
     assert_eq!(
         unsafe { crypto_generichash_init(pst, key.as_ptr(), key.len(), out.len()) },
diff --git a/libsodium-sys/tests/test.rs b/libsodium-sys/tests/test.rs
index 0c595fe8..309bb07a 100644
--- a/libsodium-sys/tests/test.rs
+++ b/libsodium-sys/tests/test.rs
@@ -1,3 +1,5 @@
+#![deny(clippy::all)]
+
 extern crate libc;
 extern crate libsodium_sys;
 ```

Gives us: 

```
[...]

PR #344, opened 2019-06-12
Add include & lib build metadata to libsodium-sys

jean-airoldie:
Adds `DEP_SODIUM_INCLUDE` & `DEP_SODIUM_LIB` env variable to can be used by dependent crates. In my case, I use it to link against `libsodium` from a c++ library (libzmq) built via cargo.

### DIFF
--- a/libsodium-sys/.gitignore
+++ b/libsodium-sys/.gitignore
@@ -1,0 +1,30 @@
+#### IDEs ####
+### CLion ###
+# CMake
+cmake-build-debug/
+
+### IntelliJ ###
+.idea
+*.iws
+*.iml
+*.ipr
+
+### Vim ###
+.sw[a-p]
+.*.sw[a-p]
+Session.vim
+.netrwhist
+*~
+
+### Visual Studio Code ###
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+.history
+
+#### Languages ####
+### Rust ###
+/target/
+Cargo.lock
--- a/libsodium-sys/README.md
+++ b/libsodium-sys/README.md
@@ -1,0 +1,12 @@
+> libsodium-sys
+
+# Build output ENV Variables
+This is the possible build metadata for the crate.
+* `DEP_SODIUM_INCLUDE` is the directory which contains the `sodium.h` header.
+    It is only available if the header was installed and `SODIUM_LIB_DIR` was not set.
+* `DEP_SODIUM_LIB` is the directory containing the compiled library.
+    It is only available if `SODIUM_LIB_DIR` was not set.
+
+See [`link build metadata`] for more information about build metadata.
+
+[`link build metadata`]: https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key
--- a/libsodium-sys/build.rs
+++ b/libsodium-sys/build.rs
@@ -6,15 +6,14 @@ extern crate libc;
 #[cfg(target_env = "msvc")]
 extern crate vcpkg;
 
-#[cfg(not(target_env = "msvc"))]
 extern crate libflate;
-
-#[cfg(not(target_env = "msvc"))]
+extern crate pkg_config;
 extern crate tar;
 
-extern crate pkg_config;
-
-use std::env;
+use std::{
+    env,
+    path::{Path, PathBuf},
+};
 
 static VERSION: &'static str = "1.0.18";
 
@@ -97,12 +96,18 @@ This function will set `cargo` flags.
 #[cfg(target_env = "msvc")]
 fn find_libsodium_pkg() {
     match vcpkg::probe_package("libsodium") {
-        Ok(_) => {
+        Ok(lib) => {
             println!(
                 "cargo:warning=Using unknown libsodium version.  This crate is tested against \
                  {} and may not be fully compatible with other versions.",
                 VERSION
             );
+            for lib_dir in &lib.link_paths {
+                println!("cargo:lib={}", lib_dir.to_str().unwrap());
+            }
+            for include_dir in &lib.include_paths {
+                println!("cargo:include={}", include_dir.to_str().unwrap());
+            }
         }
         Err(e) => {
             panic!(format!("Error: {:?}", e));
@@ -124,6 +129,12 @@ fn find_libsodium_pkg() {
                     lib.version, VERSION, lib.version
                 );
             }
+            for lib_dir in &lib.link_paths {
+                println!("cargo:lib={}", lib_dir.to_str().unwrap());
+            }
+            for include_dir in &lib.include_paths {
+                println!("cargo:include={}", include_dir.to_str().unwrap());
+            }
         }
         Err(e) => {
             panic!(format!("Error: {:?}", e));
@@ -131,130 +142,16 @@ fn find_libsodium_pkg() {
     }
 }
 
-#[cfg(any(windows, target_env = "msvc"))]
-fn get_crate_dir() -> String {
-    env::var("CARGO_MANIFEST_DIR").unwrap()
-}
-
-#[cfg(target_env = "msvc")]
-fn is_release_profile() -> bool {
-    env::var("PROFILE").unwrap() == "release"
-}
-
-#[cfg(all(target_env = "msvc", target_pointer_width = "32"))]
-fn build_libsodium() {
-    println!("cargo:rustc-link-lib=static=libsodium");
-    if is_release_profile() {
-        println!(
-            "cargo:rustc-link-search=native={}/msvc/Win32/Release/v140/",
-            get_crate_dir()
-        );
-    } else {
-        println!(
-            "cargo:rustc-link-search=native={}/msvc/Win32/Debug/v140/",
-            get_crate_dir()
-        );
-    }
-}
-
-#[cfg(all(target_env = "msvc", target_pointer_width = "64"))]
-fn build_libsodium() {
-    println!("cargo:rustc-link-lib=static=libsodium");
-    if is_release_profile() {
-        println!(
-            "cargo:rustc-link-search=native={}/msvc/x64/Release/v140/",
-            get_crate_dir()
-        );
-    } else {
-        println!(
-            "cargo:rustc-link-search=native={}/msvc/x64/Debug/v140/",
-            get_crate_dir()
-        );
-    }
-}
-
-#[cfg(all(windows, not(target_env = "msvc"), target_pointer_width = "32"))]
-fn build_libsodium() {
-    println!("cargo:rustc-link-lib=static=sodium");
-    println!(
-        "cargo:rustc-link-search=native={}/mingw/win32/",
-        get_crate_dir()
-    );
-}
-
-#[cfg(all(windows, not(target_env = "msvc"), target_pointer_width = "64"))]
-fn build_libsodium() {
-    println!("cargo:rustc-link-lib=static=sodium");
-    println!(
-        "cargo:rustc-link-search=native={}/mingw/win64/",
-        get_crate_dir()
-    );
+#[cfg(windows)]
+fn make_libsodium(_: &str, _: &Path, _: &Path) -> PathBuf {
+    // We don't build anything on windows, we simply linked to precompiled
+    // libs.
+    get_lib_dir()
 }
 
 #[cfg(not(windows))]
-fn get_archive(filename: &str) -> std::io::Cursor<Vec<u8>> {
-    use std::fs::File;
-    use std::io::{BufReader, Read};
-
-    let f = File::open(filename).expect(&format!("Failed to open {}", filename));
-    let mut reader = BufReader::new(f);
-    let mut content = Vec::new();
-    reader
-        .read_to_end(&mut content)
-        .expect(&format!("Failed to read {}", filename));
-
-    std::io::Cursor::new(content)
-}
-
-#[cfg(not(windows))]
-fn get_install_dir() -> String {
-    env::var("OUT_DIR").unwrap() + "/installed"
-}
-
-#[cfg(not(windows))]
-fn build_libsodium() {
-    use libflate::gzip::Decoder;
-    use std::process::Command;
-    use std::{fs, path::Path, str};
-    use tar::Archive;
-
-    // Determine build target triple
-    let target = env::var("TARGET").unwrap();
-
-    // Determine filenames
-    let basename = format!("libsodium-{}", VERSION);
-    let filename = format!("{}.tar.gz", basename);
-
-    // Determine source and install dir
-    let mut install_dir = get_install_dir();
-    let mut source_dir = env::var("OUT_DIR").unwrap() + "/source";
-
-    // Avoid issues with paths containing spaces by falling back to using a tempfile.
-    // See https://github.com/jedisct1/libsodium/issues/207
-    if install_dir.contains(" ") {
-        let fallback_path = "/tmp/".to_string() + &basename + "/" + &target;
-        install_dir = fallback_path.clone() + "/installed";
-        source_dir = fallback_path.clone() + "/source";
-        println!(
-            "cargo:warning=The path to the usual build directory contains spaces and hence \
-             can't be used to build libsodium.  Falling back to use {}.  If running `cargo \
-             clean`, ensure you also delete this fallback directory",
-            fallback_path
-        );
-    }
-
-    // Create directories
-    fs::create_dir_all(&install_dir).unwrap();
-    fs::create_dir_all(&source_dir).unwrap();
-
-    // Get sources
-    let compressed_file = get_archive(&filename);
-
-    // Unpack the tarball
-    let gz_decoder = Decoder::new(compressed_file).unwrap();
-    let mut archive = Archive::new(gz_decoder);
-    archive.unpack(&source_dir).unwrap();
-    source_dir.push_str(&format!("/{}", basename));
+fn make_libsodium(target: &str, source_dir: &Path, install_dir: &Path) -> PathBuf {
+    use std::{process::Command, str};
 
     // Decide on CC, CFLAGS and the --host configure argument
     let build = cc::Build::new();
@@ -349,7 +246,7 @@ fn build_libsodium() {
     }
 
     // Run `./configure`
-    let prefix_arg = format!("--prefix={}", install_dir);
+    let prefix_arg = format!("--prefix={}", install_dir.to_str().unwrap());
     let mut configure_cmd = Command::new("./configure");
     if !compiler.is_empty() {
         configure_cmd.env("CC", &compiler);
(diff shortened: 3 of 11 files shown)


kpp:
Would you please describe your use case?

jean-airoldie:
I have a crate [zeromq-src](https://github.com/jean-airoldie/zeromq-src-rs) which allows building and linking of libzmq which is a c++ library. I allow users to optionally compile the lib with `libsodium`.

To do so, I have to link against `libsodium` at the time of the c++ combilation, which means I need the directory containing the header and the compiled `libsodium` library.

Adding `DEP_SODIUM_INCLUDE` and `DEP_SODIUM_LIB` allows me to retrieve that information at the time of building. This type of build metadata is described [here](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key), and is relatively common for `sys` crates.

jean-airoldie:
Here is an example from a [mainstream crate](https://github.com/rust-lang/libz-sys/blob/master/build.rs#L155).

jean-airoldie:
Add I think adding a note about `DEP_SODIUM_INCLUDE` and `DEP_SODIUM_LIB` to a README would be ideal.

jean-airoldie:
Also this won't work on windows. I'll try to fix that.

jean-airoldie:
Hum. I'm afraid I can't do much for windows since I have no clue where the directory tree looks like. And I don't have access to a windows machine.

kpp:
> I have a crate zeromq-src which allows building and linking of libzmq which is a c++ library. I allow users to optionally compile the lib with libsodium.

I see. Let us think about it a little bit to solve your issue.

libsodium-sys/build.rs
```diff
@@ -427,4 +427,7 @@ fn build_libsodium() {
 
     println!("cargo:rustc-link-lib=static=sodium");
     println!("cargo:rustc-link-search=native={}/lib", install_dir);
+    // Accessible externally via `DEP_SODIUM_INCLUDE` env.
```
kpp: Are you sure this is enough to set env DEP_SODUIM_INCLUDE and DEP_ZMQ_LI**D**?

jean-airoldie: For my use case this is enough.

But if we wanted to be rigorous we could add `DEP_SODIUM_ROOT` which would hold `OUT_DIR` and maybe `DEP_SODIUM_PKG_CONFIG` which would point to the pkg-config file.

kpp: I am not sure you can set env var using println! in build.rs

libsodium-sys/README.md
```diff
@@ -0,0 +1,6 @@
+> libsodium-sys
+
+# Unix output ENV Variables
+These are the output ENV variables of the cargo build script:
+* `DEP_ZMQ_INCLUDE` is the directory which contains the `sodium.h` header.
```
kpp: ZMQ?

jean-airoldie: Sorry typo. I copy pasted.

libsodium-sys/build.rs
jean-airoldie: I think its quite old actually. I'll try to find the actual PR where this feature was implemented.

libsodium-sys/README.md
```diff
@@ -0,0 +1,6 @@
...
+* `DEP_ZMQ_LID` is the directory containing the compiled library
```
kpp: ZMQ_LI**D**?

libsodium-sys/build.rs
kpp: I see, thanks.

jean-airoldie: Couldn't find the PR, but based on the git blame of `libz-sys`, I know its a least 4 years old. But its quite a underground feature.

AxelNennker:
One use case is described here https://github.com/hyperledger/indy-sdk/pull/1672
Hyperledger Indy has sodiumoxide and zmq as dependencies. Both added support to compile from source using the 'vendored' feature. Now I have some functions implemented twice.

ZMQ brings its own implementation in tweetnacl.c but allows its build system to be configured to use a provided libsodium lib and include dir. @jean-airoldie zeromq-src-rs now has the feature not to use tweetnacl but needs lib and include from somewhere.

I am using the vendored feature to cross-compile libindy for Android.

kpp:
> Also this won't work on windows. I'll try to fix that.

Please mark your PR as WIP until fixed.

libsodium-sys/build.rs
```diff
@@ -427,4 +427,6 @@ fn build_libsodium() {
 
     println!("cargo:rustc-link-lib=static=sodium");
     println!("cargo:rustc-link-search=native={}/lib", install_dir);
+    println!("cargo:include={}/include", install_dir);
```
kpp: Please add comments like: "To export env:DEP_SODIUM_INCLUDE", because one may delete these lines in the future, and our CI won't catch it.

dnaq:
Can’t we add a test for it in CI?

kpp:
> Can’t we add a test for it in CI?

I believe yes

jean-airoldie:
Sure I'll add a minimal test.

As for windows support, I guess I'll figure it out using the CI, but that will be quite a pain.

libsodium-sys/build.rs
jean-airoldie: If I add a CI test that shouldn't be a problem. I can still add a comment though.

kpp: No need if you will add a test

dnaq:
It’s never bad to have comments detailing what the code does. It’s not
immediately apparent that those lines set those specific environment
variables.

jean-airoldie:
> It’s never bad to have comments detailing what the code does. It’s not immediately apparent that those lines set those specific environment variables.

Agreed but the comment might be repeated for each different compiler.

testcrate/build.rs
```diff
@@ -0,0 +1,26 @@
...
+    assert!(
+        header.is_some(),
+        "sodium.h not found in DEP_SODIUM_INCLUDE dir"
+    );
+
+    let lib = env::var("DEP_SODIUM_LIB").unwrap();
+    let compiled_lib = fs::read_dir(lib)
+        .unwrap()
+        .filter_map(Result::ok)
+        .find(|e| e.file_name() == "libsodium.a");
```
kpp: not only libsodium.a but .dll or .lib too.

testcrate/.gitignore
```diff
@@ -0,0 +1,30 @@
+#### IDEs ####
```
kpp: Do we need this file?

libsodium-sys/.gitignore
```diff
@@ -0,0 +1,30 @@
+#### IDEs ####
```
kpp: Do we need this file?

jean-airoldie: Its not strictly needed, but its practical when you are `git add .` from a subdirectory as opposed to the crate's root. I can remove it if you prefer.

testcrate/build.rs
jean-airoldie: Yes, I was waiting for the CI to fail so that I can see which file is needed in the windows build.

libsodium-sys/.gitignore
jean-airoldie: Its not an issue that you encounter if you use an IDE for instance.

jean-airoldie:
Alright so I give up for now. I can't get the windows build to work since we have to rename the `libsodium.lib` output to `sodium.lib`. However, the way the build script is setup, the file does not yet exist. Fixing this is too time consuming for me.

kpp:
> However, the way the build script is setup, the file does not yet exist.

Which one build script?

jean-airoldie:
The `libsodium-sys` build script. Are there multiple build scripts?

jean-airoldie:
I think I don't quite understand how the windows compilation work for this crate. Normally, the lib would be compiled in `build.rs` but that only seems to be the case for `unix` systems. The `windows` build script only links to a not yet existing library.

kpp:
> I think I don't quite understand how the windows compilation work for this crate.

We don't compile libsodium for the win platform, we keep precompiled binaries inside https://github.com/sodiumoxide/sodiumoxide/tree/master/libsodium-sys/msvc and https://github.com/sodiumoxide/sodiumoxide/tree/master/libsodium-sys/mingw. Also we don't download the source code, we unpack it: https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/libsodium-1.0.18.tar.gz

AxelNennker:
How are the include files found?
This PR is about building libsodium from source and setting "cargo:include={}" and "cargo:lib={}", I think.

If https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/build.rs does not support building from source for windows then this PR should not have to care about windows support.

Should e.g. https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/build.rs#L145
really be called `fn build_libsodium()` if it is not building libsodium from source? 
How about windows versions of `fn find_libsodium_env()`?
https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/build.rs#L72 
Those versions would then set `"cargo:rustc-link-search=native={}/msvc/Win32/Release/v140/"` or "cargo:rustc-link-search=native={}/msvc/Win32/Debug/v140/" or whatever.

If a later PR really builds libsodium from source in Windows that should be in the Windows versions of build_libsodium, I think.

kpp:
> How are the include files found?

We don't need headers, we unpack the source code and call ./configure + make.

> If https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/build.rs does not support building from source for windows then this PR should not have to care about windows support.

Why not? We don't build it from sources for windows because it is really painful. The easiest way was to include precompiled binaries. You can try to change the behavior but right now it is how it works.

> How about windows versions of fn find_libsodium_env()?

No, because we call find_libsodium_env only when SODIUM_DIR is set: https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/build.rs#L51

> really be called fn build_libsodium() if it is not building libsodium from source?

I don't think so. You can change the name. But you also should keep in mind we call build_libsodium in https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/build.rs#L65

AxelNennker:
This PR is about providing information to upper crates where headers and the library can be found.
On Unix where build_libsodium really builds from source the two places where headers and the library can be found by upper crates are known and provided by this PR.

[libindy](https://github.com/hyperledger/indy-sdk/) depends on

1. sodiumoxide---libsodium-sys (vendored) >>> libsodium.a + headers
2. [zmq](https://github.com/erickt/rust-zmq) (needs to find libsodium.a + headers provided by libsodium-sys (sodiumoxid)) because libzmq is build from source as well

zmq is using @jean-airoldie 's [zeromq-src-rs](https://github.com/jean-airoldie/zeromq-src-rs) crate to build zmq from source with [msvc](https://github.com/jean-airoldie/zeromq-src-rs/blob/master/src/lib.rs#L214) support.
https://github.com/erickt/rust-zmq/blob/master/zmq-sys/build/vendored.rs

In libindy I currently have the [issue](https://github.com/hyperledger/indy-sdk/pull/1672) that libsodium.a is build and linked twice. 
It would be great if libsodium-sys alone did the building and the zeromq-src-rs would use that.

Thank you for sodiumoxide and libsodium-sys

kpp:
> On Unix where build_libsodium really builds from source the two places where headers and the library can be found by upper crates are known and provided by this PR.

There is no need to build from sources to provide headers + lib. You may try to unwrap sodiumoxide/libsodium-sys/libsodium-*.tar.gz and set up env vars on windows too.

AxelNennker:
I hope @jean-airoldie can convince himself to give the Windows stuff another try.

kpp:
I will finish this PR by the end of next week.

jean-airoldie:
> > I think I don't quite understand how the windows compilation work for this crate.
> 
> We don't compile libsodium for the win platform, we keep precompiled binaries inside https://github.com/sodiumoxide/sodiumoxide/tree/master/libsodium-sys/msvc and https://github.com/sodiumoxide/sodiumoxide/tree/master/libsodium-sys/mingw. Also we don't download the source code, we unpack it: https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/libsodium-1.0.18.tar.gz

Gotcha. 

However, can't say I'm a big fan of using a tarball as opposed to a git submodule.

kpp:
> However, can't say I'm a big fan of using a tarball as opposed to a git submodule.

Previously we downloaded source code via reqwest from https://download.libsodium.org. I don't like tarballs either, but a user will have to install libgit2 and ssl, plus you have to compile a lot of deps for git2-rs to only fetch the source code. It will add 30+ deps.

jean-airoldie:
Why not simply use a git submodule? [Take this for example](https://github.com/jean-airoldie/zeromq-src-rs).

kpp:
`cargo` allows you to compile the project with multiple profiles at the same time (debug+release) so you will have to copy the submodule's directory somewhere into target/, which is painful too.

jean-airoldie:
Why would you need to do that? I'm not sure I'm following. Why wouildn't you simply output the compiled lib to the `OUT_DIR`?

kpp:
> Why wouildn't you simply output the compiled lib to the OUT_DIR?

It's not that easy because `./configure` changes local files =)

AxelNennker:
git submodules are recomended here too. https://kornel.ski/rust-sys-crate
Yes, different build need copy the source into their OUT_DIR/something. Otherwise the output of ./configure would be mixed

jean-airoldie:
> > Why wouildn't you simply output the compiled lib to the OUT_DIR?
> 
> It's not that easy because `./configure` changes local files =)

Gotcha. I thought you manually compiled using the `cc` crate.

kpp:
> Yes, different build need copy the source into their OUT_DIR/something.

I know. However you will have to copy the sources atomically or track whether you copied them or not (^C while running the build). So we gave up and simply included tarball.

AxelNennker:
why atomically?
on source copy goes to ls target/aarch64-linux-android/debug/build/libsodium-sys-4ee477c867c47cd0/out/source/
the other goes to target/aarch64-linux-android/release/build/libsodium-sys-ddfd5b6514503974/out/source/

AxelNennker:
If you did a ctrl-c a `cargo clean` helps although I think it is not required for a successful build after ctrl-c

jean-airoldie:
If you indeed need atomicity, you could compare the checksum of the `target` directory with the one from the submodule. If they don't match, then you copy. There must be is a lightweight checksum crate out there.

kpp:
> If you indeed need atomicity, you could compare the checksum of the target directory with the one from the submodule. If they don't match, then you copy. There must be is a lightweight checksum crate out there.

And you invented another one way how to make things more complicated instead of putting tarball into source code =)

jean-airoldie:
Something like [diff-dir](https://github.com/assert-rs/dir-diff) would work.

jean-airoldie:
Yeah, I guess tarball works so no need to improve.

jean-airoldie:
Not that it matters, but there probably is a git command for atomically cloning a directory.

jean-airoldie:
Alright, so 2 out the 4 windows builds pass now. I'll create a separate test PR to make sure the issues are caused by my additions.

jean-airoldie:
One of the 2 failing test is not due to this PR, and the other is for the pkg_config setup.

How should we handle the case where pkg_config is used? Are we guaranteed that a header exist along side the binaries?

kpp:
```
failures:
---- sodium_bindings::bindgen_test_layout_randombytes_implementation stdout ----
thread 'sodium_bindings::bindgen_test_layout_randombytes_implementation' panicked at 'assertion failed: `(left == right)`
  left: `24`,
 right: `48`: Size of: randombytes_implementation', libsodium-sys\src\sodium_bindings.rs:2959:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys_common::backtrace::print
```

This is not your fault. Some of -sys tests are really buggy. https://github.com/sodiumoxide/sodiumoxide/issues/338

jean-airoldie:
I handled the case we pkg_config / vcpkg is used. But I really can't do anything for the last case, when only "SODIUM_LIB_DIR" is provided. Indeed we are not guaranteed to have access to a header file.

kpp:
> Indeed we are not guaranteed to have access to a header file.

So you don't have run `cargo test -p testcrate`. And in `testcrate` to have to check whether `SODIUM_LIB_DIR` and pass the test silently.

In readme you should add a text that `DEP_` env vars are not set if the build mode is with env `SODIUM_LIB_DIR`

kpp:
> I handled the case we pkg_config / vcpkg is used. But I really can't do anything for the last case,

Then we have a really good CI set up =)

jean-airoldie:
@kpp So what blocking this PR?

kpp:
> So what blocking this PR?

CI

jean-airoldie:
CI will never pass until #338 is fixed.
> ```
> failures:
> ---- sodium_bindings::bindgen_test_layout_randombytes_implementation stdout ----
> thread 'sodium_bindings::bindgen_test_layout_randombytes_implementation' panicked at 'assertion failed: `(left == right)`
>   left: `24`,
>  right: `48`: Size of: randombytes_implementation', libsodium-sys\src\sodium_bindings.rs:2959:5
> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
> stack backtrace:
>    0: std::sys_common::backtrace::print
> ```
> 
> This is not your fault. Some of -sys tests are really buggy. #338

jean-airoldie:
@kpp Should be good now.

kpp:
Nice! Thanks

libsodium-sys/README.md
```diff
@@ -0,0 +1,10 @@
+> libsodium-sys
+
+# Build output ENV Variables
+This is the build metadata for the crate. However, note that there is no such
+metadata if `SODIUM_LIB_DIR` is set.
+* `DEP_SODIUM_INCLUDE` is the directory which contains the `sodium.h` header.
```
kpp: Is this var set always?

jean-airoldie: yes it should be always set

libsodium-sys/build.rs
```diff
@@ -6,19 +6,19 @@ extern crate libc;
...
-use std::env;
+use std::{
+    env,
+    path::{Path, PathBuf},
+};
 
 static VERSION: &'static str = "1.0.18";
 
 fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
```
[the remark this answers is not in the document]
jean-airoldie: Basically it makes it so that cargo is only rerun if build.rs is changed as opposed to any file in the root.
https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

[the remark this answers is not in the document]
jean-airoldie: But that being said, its probably not the behaviour we want so ill remove it.

libsodium-sys/README.md
kpp: Even if you use system lib without headers?

jean-airoldie: Unless `SODIUM_LIB_DIR` is set, as specified. That's because we cannot be sure that there are installed headers on the machine.

jean-airoldie: Weird, just saw your second reply. Yes indeed if pkg_config or vcpkg don't find headers this won't be set. I'll fix that.

libsodium-sys/build.rs
```diff
@@ -97,12 +96,18 @@ This function will set `cargo` flags.
...
     match vcpkg::probe_package("libsodium") {
-        Ok(_) => {
+        Ok(lib) => {
             println!(
                 "cargo:warning=Using unknown libsodium version.  This crate is tested against \
                  {} and may not be fully compatible with other versions.",
                 VERSION
             );
+            for lib_dir in &lib.link_paths {
+                println!("cargo:lib={}", lib_dir.to_string_lossy());
```
kpp: You will silently get an invalid path in case there is an invalid unicode char in the path. See https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.to_string_lossy . I believe you should use https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.to_str and panic if it returns None (invalid unicode).

libsodium-sys/build.rs
```diff
@@ -349,7 +246,7 @@ fn build_libsodium() {
     }
 
     // Run `./configure`
-    let prefix_arg = format!("--prefix={}", install_dir);
+    let prefix_arg = format!("--prefix={}", install_dir.to_string_lossy());
```
kpp: Why to_string_lossy?

libsodium-sys/build.rs
jean-airoldie: But is it actually possible to get a unicode char in these situations?

jean-airoldie: I'll make the modification, but I don't think it is possible since we don't directly get user input in these cases.

kpp:
bors: r+

kpp:
bors: r+

PR #346, opened 2019-06-14
Update CI to test all crates

jean-airoldie:
This will run `cargo test` for all packages in the workspace, including `libsodium-sys`. Note that currently some CI tests are not passing.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -44,7 +44,6 @@ cache:
   - target
 
 build: false
-
 test_script:
-  - cargo test --target %TARGET%
-  - cargo test --release --target %TARGET%
+  - cargo test --all --target %TARGET%
+  - cargo test --all --release --target %TARGET%
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,9 +44,8 @@ matrix:
       - cargo coveralls --exclude-pattern "$PWD/target/,$PWD/libsodium-sys/src/sodium_bindings.rs"
 
 script:
-  - cargo build --verbose
-  - cargo test --verbose --all
-  - cargo test --verbose --no-default-features --features std
+  - cargo test --all --verbose
+  - cargo test --all --verbose --no-default-features --features std
   - cargo doc
 
 branches:


.appveyor.yml
```diff
@@ -40,9 +44,13 @@ cache:
...
+  - cargo build --all --release --target %TARGET%
 
 test_script:
-  - cmd: cargo test --target %TARGET%
-  - cmd: cargo test --release --target %TARGET%
+  - cargo test --all --target %TARGET%
+  - cargo test --all --release --target %TARGET%
+
+on_failure:
+  - tree target /F /A
```
kpp: /F /A?

jean-airoldie: Sorry the tree command was only helpfull when I was trying to figure out the windows build layout.

.appveyor.yml
```diff
@@ -39,10 +43,8 @@ cache:
   - C:\Users\appveyor\.cargo\registry
   - target
 
-build_script:
```
kpp: Although `cargo test` builds the project, I think we should separate build and test stages.

jean-airoldie: What's the rationale?

jean-airoldie: From what I understand, for certain windows systems this helps mitigate the "directory does not contain a project or solution file" error. But my source is from 3years ago.

kpp: Because we separated them in Travis config too, whereas you merged them in Appveyor only. Also you don't have to call `cargo test --all` to test `testcrate` (with env vars), you can call it directly: `cargo test -p testcrate` IIRC.

jean-airoldie: Yes so you would prefer:
```
cargo test -p testcrate
cargo test -p libsodium-sys
cargo test -p sodiumoxyde
```
over
```
cargo test --all
```

jean-airoldie: And I can do the same for the travis build.

jean-airoldie:
I'll closed this PR. Anyway it was only to determine whether #344 caused the CI to fail. Turns out it wasn't the case #338.

kpp:
Hi! You may try to reopen this PR because of https://github.com/sodiumoxide/sodiumoxide/pull/356

Dylan-DPC:
@kpp you can't reopen the PR once the branch is deleted

kpp:
@jean-airoldie rebase onto latest master & squash commits

jean-airoldie:
@kpp sorry about that, rebased on my fork instead of upstream

jean-airoldie:
`sodium_bindings::bindgen_test_layout_randombytes_implementation` is still failing

kpp:
I see. I will try to figure out what’s going on.

adamreichold:
> > `sodium_bindings::bindgen_test_layout_randombytes_implementation` is still failing
>
> I see. I will try to figure out what’s going on.

The test failure
```
thread 'sodium_bindings::bindgen_test_layout_randombytes_implementation' panicked at 'assertion failed: `(left == right)`
  left: `24`,
 right: `48`: Size of: randombytes_implementation', libsodium-sys\src\sodium_bindings.rs:2959:5
```
seems to be rooted in `struct randombytes_implementation` being six pointers large, i.e. 24 bytes on 32-bit architectures and 48 bytes on 64-bit architectures. I guess the takeaway is that the layout test generated by `bindgen` are usually not portable even if the bindings themselves are.

I'd suggest to make this work to disable generating the layout consistency tests using `--no-layout-tests`, e.g. #404

adamreichold:
@jean-airoldie I could make sense to rebase and retry now that the bindgen layout tests are removed.

kpp:
Oops. Seems we broke tests in testcrate.
Can you fix it? We don't have to unpack libsodium anymore, we included the sources as a git submodule, so you can find version.h in libsodium-sys/libsodium/builds/msvc/version.h

jean-airoldie:
I'll take a look at it when I get the time.

adamreichold:
> Oops. Seems we broke tests in testcrate.
> Can you fix it? We don't have to unpack libsodium anymore, we included the sources as a git submodule in #393, so you can find version.h in libsodium-sys/libsodium/builds/msvc/version.h

I think the problem is that the `xcopy` invocation does work as expected (possibly because it adds the directory name to the destination path).  We would not have noticed that in #393 because the copied source is not used for building on Windows, but indeed the include meta-data points into the copied data instead of the submodule and this seems to be broken on Windows only (just tested it again on Linux where `DEP_SODIUM_INCLUDE` points at the correct directly and the test crate builds).

I'd like to add that the unpacking or copying of the source is completely unnecessary when we link the prebuilt libraries which is why #406 separates the link-against-prebuilt and build-from-source approaches and sets `DEP_SODIUM_INCLUDE` to just point into the submodule in the first case which should fix this AFAIU.

adamreichold:
> I think the problem is that the xcopy invocation does work as expected

Indeed that was one of the two problems. I took the commit from here and added another one to fix this in #407

kpp:
Closed in favor of https://github.com/sodiumoxide/sodiumoxide/pull/407

@jean-airoldie thanks! Sorry for the delay.

Issue #349, opened 2019-07-09
replace uses of `std::mem::uninitialized` with `MaybeUninit`


dnaq:
`std::mem::uninitalized` has been deprecated in the latest rust compiler (for good reasons). We should replace all uses of it with `MaybeUninit` or with explicit initialization before the next release.

kpp:
```
src/crypto/sign/ed25519.rs
177:            let mut s = mem::uninitialized();

src/crypto/generichash/mod.rs
59:            state = mem::uninitialized();

src/crypto/secretstream/xchacha20poly1305.rs
48:        let ciphertext: [u8; (ABYTES - 1)] = unsafe { mem::uninitialized() };

src/crypto/secretstream/secretstream_macros.rs
172:        let mut header: [u8; HEADERBYTES] = unsafe { mem::uninitialized() };
174:            unsafe { mem::uninitialized() };
249:            unsafe { mem::uninitialized() };
300:        let mut tag: u8 = unsafe { mem::uninitialized() };

src/crypto/auth/auth_state_macros.rs
51:            let mut s = mem::uninitialized();

src/crypto/hash/hash_macros.rs
41:            let mut st: $hash_state = mem::uninitialized();
```

PR #350, opened 2019-07-17
Replace mem::uninitialized with MaybeUninit

kpp:
Closes https://github.com/sodiumoxide/sodiumoxide/issues/349

### DIFF
--- a/src/crypto/secretstream/secretstream_macros.rs
+++ b/src/crypto/secretstream/secretstream_macros.rs
@@ -169,19 +169,22 @@ impl Stream<Push> {
     /// encryption applications can use the password hashing API to get a key
     /// that can be used with the functions below.
     pub fn init_push(key: &Key) -> Result<(Stream<Push>, Header), ()> {
-        let mut header: [u8; HEADERBYTES] = unsafe { mem::uninitialized() };
-        let mut state: $state_name =
-            unsafe { mem::uninitialized() };
+        let mut header = mem::MaybeUninit::<[u8; HEADERBYTES]>::uninit();
+        let mut state = mem::MaybeUninit::uninit();
         let rc = unsafe {
             $init_push_name(
-                &mut state,
-                header.as_mut_ptr(),
+                state.as_mut_ptr(),
+                header.as_mut_ptr() as *mut u8,
                 key.0.as_ptr(),
             )
         };
         if rc != 0 {
             return Err(());
         }
+        // rc == 0 and both state and header are initialized
+        let state = unsafe { state.assume_init() };
+        let header = unsafe { header.assume_init() };
+
         Ok((
             Stream::<Push> {
                 state,
@@ -245,11 +248,10 @@ impl Stream<Pull> {
     /// will not be required any more for subsequent operations. `Err(())` is
     /// returned if the header is invalid.
     pub fn init_pull(header: &Header, key: &Key) -> Result<Stream<Pull>, ()> {
-        let mut state: $state_name =
-            unsafe { mem::uninitialized() };
+        let mut state = mem::MaybeUninit::uninit();
         let rc = unsafe {
             $init_pull_name(
-                &mut state,
+                state.as_mut_ptr(),
                 header.0.as_ptr(),
                 key.0.as_ptr(),
             )
@@ -263,6 +265,9 @@ impl Stream<Pull> {
         } else if rc != 0 {
             return Err(());
         }
+        // rc == 0 and state is initialized
+        let state = unsafe { state.assume_init() };
+
         Ok(Stream::<Pull> {
             state: state,
             finalized: false,
@@ -297,7 +302,7 @@ impl Stream<Pull> {
         let (ad_p, ad_len) = ad
             .map(|ad| (ad.as_ptr(), ad.len()))
             .unwrap_or((ptr::null(), 0));
-        let mut tag: u8 = unsafe { mem::uninitialized() };
+        let mut tag = mem::MaybeUninit::uninit();
         let mut buf = Vec::with_capacity(m_len);
         let rc = unsafe {
             buf.set_len(m_len);
@@ -305,7 +310,7 @@ impl Stream<Pull> {
                 &mut self.state,
                 buf.as_mut_ptr(),
                 &mut (m_len as c_ulonglong),
-                &mut tag,
+                tag.as_mut_ptr(),
                 c.as_ptr(),
                 c_len as c_ulonglong,
                 ad_p,
@@ -315,6 +320,9 @@ impl Stream<Pull> {
         if rc != 0 {
             return Err(());
         }
+        // rc == 0 and tag is initialized
+        let tag = unsafe { tag.assume_init() };
+
         let tag = Tag::from_u8(tag)?;
         if tag == Tag::Final {
             self.finalized = true;
--- a/src/crypto/secretstream/xchacha20poly1305.rs
+++ b/src/crypto/secretstream/xchacha20poly1305.rs
@@ -45,7 +45,7 @@ mod test {
 
     #[test]
     fn decrypt_too_short_ciphertext() {
-        let ciphertext: [u8; (ABYTES - 1)] = unsafe { mem::uninitialized() };
+        let ciphertext = [0; (ABYTES - 1)];
         let key = gen_key();
         let (_, header) = Stream::init_push(&key).unwrap();
         let mut stream = Stream::init_pull(&header, &key).unwrap();
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@
 
 This package aims to provide a type-safe and efficient Rust binding that's just
 as easy to use.
+Rust >= 1.36.0 is required because of mem::MaybeUninit.
 
 ## Basic usage
 
--- a/src/crypto/auth/auth_state_macros.rs
+++ b/src/crypto/auth/auth_state_macros.rs
@@ -47,11 +47,12 @@ impl Drop for State {
 impl State {
     /// `init()` initializes an authentication structure using a secret key 'k'.
     pub fn init(k: &[u8]) -> State {
-        unsafe {
-            let mut s = mem::uninitialized();
-            $init_name(&mut s, k.as_ptr(), k.len());
-            State(s)
-        }
+        let mut s = mem::MaybeUninit::uninit();
+        let state = unsafe {
+            $init_name(s.as_mut_ptr(), k.as_ptr(), k.len());
+            s.assume_init() // s is definitely initialized
+        };
+        State(state)
     }
 
     /// `update()` can be called more than once in order to compute the authenticator
--- a/src/crypto/generichash/mod.rs
+++ b/src/crypto/generichash/mod.rs
@@ -53,18 +53,19 @@ impl State {
             }
         }
 
-        let mut state: crypto_generichash_state;
+        let mut state = mem::MaybeUninit::uninit();
 
         let result = unsafe {
-            state = mem::uninitialized();
             if let Some(key) = key {
-                crypto_generichash_init(&mut state, key.as_ptr(), key.len(), out_len)
+                crypto_generichash_init(state.as_mut_ptr(), key.as_ptr(), key.len(), out_len)
             } else {
-                crypto_generichash_init(&mut state, ptr::null(), 0, out_len)
+                crypto_generichash_init(state.as_mut_ptr(), ptr::null(), 0, out_len)
             }
         };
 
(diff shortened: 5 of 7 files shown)


src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -169,19 +169,22 @@ impl Stream<Push> {
...
-        let mut state: $state_name =
-            unsafe { mem::uninitialized() };
+        let mut header = mem::MaybeUninit::<[u8; HEADERBYTES]>::uninit();
+        let mut state = mem::MaybeUninit::uninit();
         let rc = unsafe {
             $init_push_name(
-                &mut state,
-                header.as_mut_ptr(),
+                state.as_mut_ptr(),
+                header.as_mut_ptr() as *mut u8,
```
Stargateur: Are you sure about this one ? a double `as_mut_ptr()` should avoid the cast `header.as_mut_ptr().as_mut_ptr()`. And this make me think that `[mem::MaybeUninit::<u8>; HEADERBYTES]` should make more sense ?

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -297,7 +302,12 @@ impl Stream<Pull> {
         let (ad_p, ad_len) = ad
             .map(|ad| (ad.as_ptr(), ad.len()))
             .unwrap_or((ptr::null(), 0));
-        let mut tag: u8 = unsafe { mem::uninitialized() };
+        let mut tag = {
+            // Tag can be one of Message=0, Push=1, Rekey=2, Final=3 states
+            // So even if tag is not initialized in $pull_name call
+            //  we will get an error in Tag::from_u8 which is better than UB
+            u8::max_value()
+        };
```
Stargateur: This is less unsafe but you just remove unsafe and make the code more slow. It's a choice but I think you mix the clear goal to replace `uninitialized` by `MaybeUninit` and the goal to remove unsafe at all cost. This should be in a separate PR I think.

src/crypto/secretstream/xchacha20poly1305.rs
```diff
@@ -45,7 +45,7 @@ mod test {
 
     #[test]
     fn decrypt_too_short_ciphertext() {
-        let ciphertext: [u8; (ABYTES - 1)] = unsafe { mem::uninitialized() };
+        let ciphertext = [0; (ABYTES - 1)];
```
Stargateur: This is less unsafe but you just remove unsafe and make the code more slow. It's a choice but I think you mix the clear goal to replace `uninitialized` by `MaybeUninit` and the goal to remove unsafe at all cost. This should be in a separate PR I think.

src/crypto/secretstream/secretstream_macros.rs
dnaq: I feel like using `[mem::MaybeUninit::<u8>; HEADERBYTES]` would lead to uglier code. I'm ok with the pointer cast.

src/crypto/secretstream/secretstream_macros.rs
dnaq: Agreed, just use `MaybeUninit` here.

src/crypto/secretstream/xchacha20poly1305.rs
dnaq: It's in test code, so performance is not an issue.

src/crypto/secretstream/secretstream_macros.rs
kpp: Me too.

src/crypto/secretstream/secretstream_macros.rs
kpp: Fixed

Dylan-DPC:
A concern is this requires the minimum version to be bumped to 1.36.0 which means the builds will fail for older versions. How are we planning of handling this? I'd suggest a new major release

lnicola:
There's no consensus for that and each project can decide, but personally I think major version bumps on minimum compiler version changes are unproductive:

- old versions of the Rust compiler and standard library get no security updates or support of any kind
- bumping the major version tends to split the ecosystem in two, and in the end produces much more churn across the other crates
- projects that insist on using an old compiler can lock their dependencies
- holding off code changes and dependency upgrades is ultimately worse for the users e.g.:
  - `serde` is still using `mem::uninitialized` and accepting possible UB in order to keep supporting old compilers
  - when dependencies don't want to upgrade, it's very easy to get multiple versions of the same crate in your project, which can easily balloon your build times

src/crypto/secretstream/secretstream_macros.rs
Stargateur: Something like that would look better but currently require nightly:

```rust
let mut header: [mem::MaybeUninit::<u8>; HEADERBYTES] = unsafe { mem::MaybeUninit::uninit().assume_init() };
let mut state = mem::MaybeUninit::uninit();
let rc = unsafe {
    $init_push_name(
        state.as_mut_ptr(),
        mem::MaybeUninit::first_ptr_mut(&mut header),
        key.0.as_ptr(),
    )
};
if rc != 0 {
    return Err(());
}
// rc == 0 and both state and header are initialized
let state = unsafe { state.assume_init() };
let header = unsafe { mem::transmute(header) };
```

kpp: If HEADERBYTES is small enough we can initialize header with zeroes.

> `unsafe { mem::transmute(header) };`

Please, no =)

Stargateur: That what the doc of std advice https://doc.rust-lang.org/nightly/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element

Dylan-DPC:
> @Dylan-DPC This is not a breaking change to require a higher compiler version, the only thing that could be a breaking is to going from stable to nightly https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#major-change-going-from-stable-to-nightly.
> As MaybeUninit is stable, this is ok. Just run rustup update.

@Stargateur not really. It would fail on older versions and we can't expect people using it to just update their rust versions and crates using sodiumoxide to expect their downstream users to update their rust version.

Dylan-DPC:
@lnicola those are actually inaccurate facts. Changes are often backported to older versions if deemed necessary or a major security issue. minimum rust version has nothing to do with most of it and those issues can be faced either way - even if msvr is bumped and dependencies are not updated for example

lnicola:
> Changes are often backported to older versions if deemed necessary or a major security issue.

Maybe in `sodiumoxide`, but not in the compiler and standard library, which is what I was talking about.

> those issues can be faced either way - even if msvr is bumped and dependencies are not updated for example

Indeed, they can happen. But if you break semver compat for MSRV, you practically force them to happen.

[Here's](https://github.com/rayon-rs/rayon/issues/586) an example of `rayon` missing out on improvements to dependencies because of not wanting to require a newer compiler. [Here](https://github.com/serde-rs/serde/blob/7b0e06c82569bdb72554e8ee61f1afab23a5a3e9/serde/src/ser/impls.rs#L621) is `serde` using `mem::uninitialized` for the same reason.

Dylan-DPC:
> Maybe in sodiumoxide, but not in the compiler and standard library, which is what I was talking about.

I'm talking about std libraries and rustc as well. Every time there is a bug there is always discussion on how many previous versions are affected and whether they require patches depending on the severity of the issue.

Given the choice between breaking builds with older compilers and not using newer features, we prefer to stick to the latter as we have in the past as well.

src/crypto/secretstream/secretstream_macros.rs
RalfJung: `[mem::MaybeUninit::<u8>::uninit(); HEADERBYTES]` should work to create an uninitialized array without `unsafe`; `MaybeUninit<u8>` is `Copy`.

kpp:
LGTY?

Dylan-DPC:
yeah we can merge this and release a 0.3.0 whenever ready for release

RalfJung:
Why is this breaking? Is it because it bumps the minimal Rust version?

kpp:
Yes

dnaq:
I don’t see that as breaking to be honest. The library API hasn’t changed,
and I’d prefer to not have users left on old versions of the library, or on
old compilers either tbh.

Dylan-DPC:
This is a "borderline" issue and is left to the crate authors to decide whether to consider it breaking or not. Imo anyone who hasn't updated rust would end up with a broken build (generally dodged by feature gating or build scripts that build differently based on different rust versions). I neither like older compilers nor stale old versions of the crate, but then as crate maintainers we have to often take the defensive path.

RalfJung:
What is your current minimal version? If you can tolerate an extra dependency, https://crates.io/crates/maybe-uninit can help. It doesn't actually make things sound on old versions, but it means the same code can work (unsoundly) on old versions and (soundly) on Rust 1.36+. That is assuming testing confirms that the UB in old versions is not causing problems.

kpp:
>  If you can tolerate an extra dependency, https://crates.io/crates/maybe-uninit can help.

@dnaq ?

kpp:
In my last commit https://github.com/sodiumoxide/sodiumoxide/pull/350/commits/c897829e09128ea2436e61ff6ff2dc1e954331e3 I replaced std::mem::MaybeUninit with maybe_uninit, so we no longer need a 1.36 rustc. @dnaq what do you think? Or I can revert it.

Dylan-DPC:
If it doesn't make it sound on older version than what's the point of adding a new dependency?

kpp:
I dunno. It's just an option.

kpp:
@dnaq @Dylan-DPC you guys need to come up to a conclusion about which one MaybeUninit we should use (std or maybe_uninit) and how to release a new version of the crate (minor or a patch).

dnaq:
I think we should go with the standard library and also only bump the minor
version. I’m on mobile right now, but I can elaborate on why later on.

kpp:
> and also only bump the minor

By minor you mean 0.3.0?

Dylan-DPC:
The external crate still exposes the UB for lower compiler versions so the only plus point is that people don't have to bump a major version which isn't much. So i'd prefer if we stay with the former approach (of not using the dependency).

kpp:
I removed maybe-uninit crate in favor of std

dnaq:
So. To elaborate on my reasoning. I don’t believe that we should break
semver compatibility because we require a newer compiler version.

The reason is that rust has a really good story for always running the
latest stable compiler and I believe all security critical projects should
be in the habit of doing so. The rust compiler team basically guarantees
that code written for a certain edition will always work on stable
compilers, modulo critical security fixes. I believe this should lead to
projects having methods in place for running a recent stable compiler.

Technically our APIs haven’t changed a bit by this pull request, so the
major version number shouldn’t increase (the minor in our case since we’re
still at version 0).

I’m also really wary about increasing it for the reason that we won’t have
the resources to back port any potential security issues in the future to
the old version, if we have people running it since they use old compilers
that will be an issue.

So that’s my reasoning, I might be wrong, and maybe the best way forward is
to use the maybe_uninit crate, but that would go against me wanting people
in the rust ecosystem to use recent compilers. My decision isn’t final, and
I’m open to differing arguments.

kpp:
> My decision isn’t final, and I’m open to differing arguments.

@dnaq I agree with you

Dylan-DPC:
> The reason is that rust has a really good story for always running the
latest stable compiler 

Sadly this isn't the case, a lot of dependents often depend or support older compiler versions. Then there is also the case of distro support, where in distros often support a certain version and do not update for a long time (this is changing lately). Doing a breaking release even though nothing is actually a "breaking" change is far better than doing a non-breaking release with a breaking change.

kpp:
@Dylan-DPC the users can always pin to the version that works on their compiler. We will receive several bug reports and solve them by either moving to maybe_uninit or helping them to migrate to a new version of rustc.

Dylan-DPC:
Yes, but i'm not a fan of pinning to a certain version, as this goes against semver policies. (it should be possible to install any backward compatible version and if a minor isn't backward compatible with the other versions, we are violating semver at that point, though also semver says that for 0.x versions, "minor" (or patch ) releases may not be compatible, but we generally like to keep it compatible)

kpp:
I suggest merging this PR, we can figure out how to cut a new release a little bit later.

Dylan-DPC:
Yes i was waiting to see if we want to push any changes first to 0.2.* and then merge this for 0.3.0 or whether we could change it anyway.  merging it

Dylan-DPC:
bors: r+

PR #351, opened 2019-07-20
Kdf bindings

Richard-W:
This pull request adds bindings for the `crypto_kdf_blake2b_derive_from_key` function


Richard-W:
force-push fixed an issue with the doc test

kpp:
There are several lines not covered with your tests: https://coveralls.io/builds/24694535/source?filename=crypto/kdf/blake2b.rs

Please increase the coverage

Richard-W:
I added a test for crypto::kdf::gen_key(). Forgot about that. Sorry.

I don't know any way to trigger the error in kdf/blake2b.rs:54 though. As far as I know this error is emitted by libsodium if the size of the subkey not in the `[ffi::crypto_kdf_blake2b_BYTES_MIN, ffi::crypto_kdf_blake2b_BYTES_MIN]` interval. This can never happen in this implementation so this might actually be dead code. I don't feel comfortable removing the check though.

kpp:
> I don't feel comfortable removing the check though.

Me too.

Rebase on the latest master or merge it into your branch.

Richard-W:
Done. Rebased to current master

src/crypto/kdf/blake2b.rs
```diff
@@ -0,0 +1,138 @@
...
+use ffi;
+
+/// Number of bytes in a `Key`.
+pub const KEYBYTES: usize = ffi::crypto_kdf_blake2b_KEYBYTES as usize;
+
+/// Number of bytes in a `Context`
+pub const CONTEXTBYTES: usize = ffi::crypto_kdf_blake2b_CONTEXTBYTES as usize;
+
+/// Number of bytes in a `SubKey`.
+pub const SUBKEYBYTES: usize = 32;
```
kpp: Why 32?

kpp:
Several functions are missing. See https://github.com/jedisct1/libsodium/blob/master/test/default/kdf.c

src/crypto/kdf/blake2b.rs
Richard-W: Had to make a choice here for a fixed length type. The `pwhash` module does something similar for `Salt`. 32 seemed a reasonable choice since its equal to the keysize for the secretbox and secretstream APIs.

Do you prefer a dynamic-size type?

Richard-W:
What functions do you mean? Stuff like `crypto_kdf_keybytes()` etc. is expressed as a constant (`KEYBYTES`) here like its done in all other modules. Did I miss something?

kpp:
> Did I miss something?

min/max bytes

Richard-W:
Its not missing. It got replaced by `SUBKEYBYTES` analogue to what `pwhash` does with Salt.

Thinking about it again, this may not be a good idea (and my docs are lacking). I will refactor this.

src/crypto/kdf/blake2b.rs
kpp: I am not sure how to use it. I don't not what is kdf/blake2b for.

Richard-W:
I refactored the API. It is now closer to the original libsodium API

src/crypto/kdf/blake2b.rs
Richard-W: KDF stands for key derivation function. You can use it to derive multiple independent subkeys from a single master key. This is used for example to derive rx and tx keys from a single key exchange.

Dylan-DPC:
looks good to me, would like a review from @dnaq

src/crypto/kdf/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+//! ```
+//! use sodiumoxide::crypto::kdf;
+//! use sodiumoxide::crypto::secretbox;
+//!
+//! const CONTEXT: [u8; 8] = *b"Examples";
+//!
+//! let key = kdf::gen_key();
+//!
+//! let mut key1 = secretbox::Key([0; secretbox::KEYBYTES]);
+//! kdf::derive_from_key(&mut key1.0[..], 1, CONTEXT, &key).unwrap();
```
kpp: Will it generate correct `crypto::secretbox::SecretKey`s?

Richard-W: Yes. Any array of the correct size is a valid key for secretbox

kpp:
@dnaq review plz

dnaq:
Sorry. I’m really swamped right now. Can someone else review it?

src/crypto/kdf/blake2b.rs
```diff
@@ -0,0 +1,163 @@
...
+    subkey_id: u64,
+    ctx: [u8; CONTEXTBYTES],
+    key: &Key,
+) -> Result<(), ()> {
+    unsafe {
+        let r = ffi::crypto_kdf_blake2b_derive_from_key(
+            subkey.as_mut_ptr() as _,
+            subkey.len(),
+            subkey_id,
+            ctx.as_ptr() as *mut i8,
```
dvic: Can't this just be `ctx.as_ptr()`?

Richard-W: The `*mut i8` cast is necessary since `&[u8]::as_ptr` returns `*mut u8` and rust has no notion of type compatibility in this case.

kpp: Wait a minute. https://doc.rust-lang.org/std/primitive.slice.html#method.as_ptr returnds `*const u8` you cast it to `*mut i8`. Is it legal?

GallagherCommaJack: There's no good reason for `ctx` to be a cast to a `*mut`, and it's sketchy as hell to cast it to one. The following code typechecks just fine:
```suggestion
            ctx.as_ptr() as *const i8,
```

GallagherCommaJack: For that matter so does
```suggestion
            ctx.as_ptr() as _,
```

kpp: > ctx.as_ptr() as *const i8,

Nice

kpp:
Please rebase onto latest master

src/crypto/kdf/blake2b.rs
Richard-W: I commited your suggestion. I feel slightly ashamed... stupid mistake...

kpp:
@Richard-W please rebase onto latest master

arianvp:
I have a rebased version of this branch here: https://github.com/arianvp/sodiumoxide/tree/richard-w/kdf-bindings

Richard-W:
Sorry for the delay. I am kind of swamped right now. Branch is now rebased to current master.

Richard-W:
Damn. I rebased an outdated version of the branch. Moment please

Richard-W:
rebased to current master

kpp:
bors: r+

PR #357, opened 2019-07-24
Allow reusing Vec for secretstream

### LINKED ISSUE #353: Allow reusing Vec for secretstream
The current secretstream API requires allocating a new `Vec` for every call to `push` or `pull`. Ideally, there would be a way to use an existing `Vec` (or maybe even just a `&mut byte[]`) so that a client that repeatedly calls push or pull can reuse the same allocated memory for multiple calls.

tmccombs:
Fixes #353

### DIFF
--- a/src/crypto/secretstream/secretstream_macros.rs
+++ b/src/crypto/secretstream/secretstream_macros.rs
@@ -64,7 +64,7 @@ const TAG_FINAL: u8 = $tag_final as u8;
 /// code is generated over all data. A typical encrypted stream simply attaches
 /// `0` as a tag to all messages, except the last one which is tagged as
 /// `Tag::Final`. When decrypting the tag is retrieved and may be used.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Copy, Clone)]
 pub enum Tag {
     /// Message, the most common tag, that doesn't add any information about the
     /// nature of the message.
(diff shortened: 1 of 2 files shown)


src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -196,6 +196,33 @@ impl Stream<Push> {
...
+        let mut buf = Vec::with_capacity(buf_len);
+        self.push_impl(m, ad, tag, &mut buf)?;
+        Ok(buf)
+    }
+
+    /// All data (including optional fields) is authenticated. Encrypts a
+    /// message `m` and its `tag`. Optionally includes additional data `ad`,
+    /// which is not encrypted.
+    ///
+    /// The encrypted message is written to the `out` vector, ovwerwriting any existing data there.
```
kpp: > ovwerwriting

src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -196,6 +196,33 @@ impl Stream<Push> {
...
+        let buf_len = self.push_check(m, tag)?;
+        out.clear();
+        out.reserve(buf_len);
+        self.push_impl(m, ad, tag, out)
+    }
+
+    /// Create a ciphertext for an empty message with the `TAG_FINAL` added
+    /// to signal the end of the stream. Since the `Stream` is not usable
+    /// after this point, this method consumes the `Stream.
+    pub fn finalize(mut self, ad: Option<&[u8]>) -> Result<Vec<u8>, ()> {
```
kpp: Why did you move this method from the bottom of the impl into the middle?

kpp:
Also there are several lines not covered with tests: https://coveralls.io/builds/24783823/source?filename=crypto%2Fsecretstream%2Fsecretstream_macros.rs#L221

Please increase the coverage

src/crypto/secretstream/secretstream_macros.rs
tmccombs: Because I put the private methods at the end. However, I'd be happy to move it back down to the bottom if you think that is better style.

kpp: I see. We don't have any style for putting private methods at the end.

kpp:
bors: r+

kpp:
bors: retry

PR #362, opened 2019-08-18
Added ed25519 to x25519 conversion methods

ryankurte:
Allowing use of signing keys for encryption/decryption if required, as in the [docs](https://download.libsodium.org/doc/advanced/ed25519-curve25519). This is implemented as `TryInto` on `&T` so keys are not consumed in the conversion.

I'm struggling to see if/where there are test vectors in the underlying [library](https://github.com/jedisct1/libsodium/blob/927dfe8e2eaa86160d3ba12a7e3258fbc322909c/test/default/ed25519_convert.c), so the tests are p primitive.


ryankurte:
hey hey, anything else i need to do here?

kpp:
https://github.com/sodiumoxide/sodiumoxide/pull/361#issuecomment-519459157

sekhat:
Hey guys, is there anything holding this pull request up?

dnaq:
Hi,

Sorry for the extremely long delay.

I think I might prefer the approach of exposing conversion functions instead of TryInto, even though TryInto is more idiomatic rust we don't really want people to use this functionality unless they really need it. I think TryInto might lead to people using it without considering alternative solutions.

However I'm not the maintainer anymore, so I'll do a code review as well if @kpp differs in his opinion.

src/crypto/sign/ed25519.rs
```diff
@@ -234,6 +234,58 @@ impl Default for State {
     }
 }
 
+use crate::crypto::box_::{
```
dnaq, line 237: Maybe do`use crate::crypto::box` instead of renaming the constants.

src/crypto/sign/ed25519.rs
```diff
@@ -234,6 +234,58 @@ impl Default for State {
...
+impl TryFrom<&PublicKey> for Curve25519PublicKey {
+    type Error = ();
+
+    /// Converts an Ed25519 public key `crypto::sign::PublicKey` to an X25519 public key and stores it into a `crypto::box_::PublicKey`
+    fn try_from(ed25519_pk: &PublicKey) -> Result<Curve25519PublicKey, ()> {
+        let mut x25519_pk = Curve25519PublicKey([0u8; CURVE25519_PUBLICKEYBYTES]);
+
+        let ret = unsafe {
+            ffi::crypto_sign_ed25519_pk_to_curve25519(
+                (&mut x25519_pk.0).as_mut_ptr(),
```
dnaq, line 254: This feels like double conversions. Shouldn't `pk.0.as_mut_ptr()` be enough?

src/crypto/sign/ed25519.rs
```diff
@@ -234,6 +234,58 @@ impl Default for State {
...
+                (&ed25519_pk.0).as_ptr(),
```
dnaq, line 255: Same as above

src/crypto/sign/ed25519.rs
```diff
@@ -234,6 +234,58 @@ impl Default for State {
...
+impl TryFrom<&SecretKey> for Curve25519SecretKey {
+    type Error = ();
+
+    /// Converts an Ed25519 secret key `crypto::sign::SecretKey` to an X25519 secret key and stores it into an `crypto::box_::SecretKey`
+    fn try_from(ed25519_sk: &SecretKey) -> Result<Curve25519SecretKey, ()> {
+        let mut x25519_sk = Curve25519SecretKey([0u8; CURVE25519_SECRETKEYBYTES]);
+
+        let ret = unsafe {
+            ffi::crypto_sign_ed25519_sk_to_curve25519(
+                (&mut x25519_sk.0).as_mut_ptr(),
```
dnaq, line 276: Same

src/crypto/sign/ed25519.rs
```diff
@@ -234,6 +234,58 @@ impl Default for State {
...
+                (&ed25519_sk.0).as_ptr(),
```
dnaq, line 277: Same

dnaq (changes requested):

sekhat:
I've gone ahead and implemented the changes requested here as PR #456.

PR #369, opened 2019-09-01
Fix deprecated try! macro

kpp:


dnaq:
I look forward to the day when const generics are stable so that we can replace the hand written serde implementations with derives.

kpp:
bors: r+

kpp:
bors: retry

Issue #366, opened 2019-08-30
make `prelude` dependency optional


GallagherCommaJack:
Would this even make sense? It would definitely be nice to be able to use this from systems that for some reason or another don't even have `alloc`, but I'm not sure whether or not `libsodium` relies on the same functionality as `alloc`.

From a quick glance it wouldn't be too hard to get `sodiumoxide` out of the way of that dream though - I've started going through marking things with feature flags, and the only snag I've run into is that I'm not 100% sure how to deal with e.g. `stream_module` functions returning `Vec`. Whether or not this whole endeavor makes sense I think replacing anything that uses `Vec::set_len` with functions that modify a buffer in place will do a lot to make the memory safety of `unsafe` code more obvious.

GallagherCommaJack:
All work I do on this front is going [here](https://github.com/kalix-systems/sodiumoxide/tree/no_std).

kpp:
Great! Also we got this issue: https://github.com/sodiumoxide/sodiumoxide/issues/197

GallagherCommaJack:
As of [this commit](https://github.com/kalix-systems/sodiumoxide/commit/226a3af782688c1ace4bee00e43e73fb9d0f4e5a) all tests are passing again and all libtests are passing. There are a few things that I want to fix before making this a pull request though:

1) The doctests fail without the `alloc` feature, since they rely on functions that allocate a new vector. The right thing to do is probably to conditionally include those docs, and add corresponding conditional docs for the no-`alloc` scenario.
2) There are a lot of tests that don't strictly *need* to heap allocate, but I didn't want to deal with counting the bytes in the test vectors.
3) There are other tests that need variable length arrays with an upper bound known at compile time, and so I might replace those with `ArrayVec`.

kpp:
Please open a new PR

PR #371, opened 2019-09-02
Make alloc dependency optional

GallagherCommaJack:
Related issue: #366 
 
The API has been slightly extended, in particular all operations that allocated or extended a `Vec` now offer an equivalent that operates in-place on a slice. I also replaced some functions that looked like they would operate in place (by taking an `&mut Vec<u8>` and extending it) with operations that take an `&mut [u8]` instead.

This definitely shouldn't get included without a version bump, since it can cause code that was relying on that functionality to break without causing a type error.

I'm actually somewhat inclined to deprecate a lot of the functions I had to alter in favor of functions more like `box_::open_detached` which put more of the requirements in at the type level rather than at runtime, but that would be a larger change than was easy to make.


kpp:
The question is simple: can you enable `alloc` feature without `std` and vice versa?

GallagherCommaJack:
> The question is simple: can you enable `alloc` feature without `std` and vice versa?

You can enable `alloc` without `std`, but it wouldn't really make sense to enable `std` without `alloc`, so you can't.

kpp:
Sorry for the delay. Would you please rebase the branch?

run_tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+cargo check
+cargo test
+
+cargo check --no-default-features
+cargo test --no-default-features --lib
+
+cargo check --no-default-features --features="alloc"
+cargo test --no-default-features --features="alloc" --lib
```
kpp: Maybe it would be easier to build with `--features std` and `--features alloc` but test the code only with `--features std`? It would have reduced a lot of `#[cfg(feature = "alloc")] #[cfg(test)]`.

kpp:
@GallagherCommaJack Hello. We have changed our CI a little bit. Would you please rebase you PR onto the latest master?

kpp:
I am sorry to close your PR for inactivity. I will be happy to be working on your PR when you want to continue.

PR #375, opened 2019-09-16
Create ed25519 key pair from 32 byte secret key

llowrey:
Added ability to create ed25519 key pair from 32 byte secret key slice. Libsodium's secret key is the 32 byte secret key concatenated with the 32 byte public key. Openssl PEM/DER files contain just the 32 byte secret key.

The new method keypair_from_secret_key_slice() accepts the 32 byte secret key and derives the 32 byte public key using scalarmult_base. It then creates a new SecretKey using the resulting 64 byte concatenation of the secret and public keys. It returns the resulting SecretKey and the PublicKey.

### DIFF
--- a/src/crypto/scalarmult/ed25519.rs
+++ b/src/crypto/scalarmult/ed25519.rs
@@ -1,0 +1,129 @@
+//! `crypto_scalarmult_curve25519` specified in
+//! [Cryptography in `NaCl`](http://nacl.cr.yp.to/valid.html), Sections 2, 3, and 4.
+//! This function is conjectured to be strong. For background see Bernstein,
+//! "Curve25519: new Diffie-Hellman speed records," Lecture Notes in Computer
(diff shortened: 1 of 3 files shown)


dnaq:
A very nicely made pull request (containing tests and everything), but how does this differ from the `keypair_from_seed()` function that we already expose?

llowrey:
facepalm

When I originally looked at `crypto_sign_ed25519_seed_keypair` I misinterpreted what it was doing and assumed that it was generating both the sk and pk from the seed in a nacl/libsodium proprietary manner. I now see that it is just implementing standard rfc8032 key generation. That method could use a better name, imo.

Sorry for the noise.

kpp:
@llowrey does it mean that your PR does not add a new functionality?

kpp:
Closed due to inactivity. @llowrey please file a new PR if you think your PR brings new functionality.

PR #378, opened 2019-10-06
Use `copy_from_slice` instead of manual loop where possible

AnthonyMikh:
There is no need to reinvent the wheel.


kpp:
bors: r+

kpp:
bors: retry

Issue #377, opened 2019-09-18
libsodium compilation. We do it wrong.


kpp:
It all started when I tried to crosscompile sodiumoxide to wasm32-wasi.

```
#[cfg(all(target_arch="wasm32", target_os="wasi"))]
fn build_libsodium() {
    std::compile_error!("wasm32")
}

#[cfg(target_arch = "")] // or comment out this function
fn build_libsodium() {
```

Did not help at all. I kept receiving error:

```
error[E0425]: cannot find function `build_libsodium` in this scope
  --> libsodium-sys/build.rs:64:9
   |
64 |         build_libsodium();
   |         ^^^^^^^^^^^^^^^ not found in this scope
```

The issue is: build.rs gets compiled on host. So our target_ predicates in build.rs are equal to host target_ predicates.
In order to make everything work perfectly we have to:

```
cross_compilation = env:HOST != env:TARGET 

if (!cross_compilation) {
    if on windows {
        try find libsodium from vcpkg
        try find libsodium in env vars
        or unpack prebuilt library e.g. msvc/x64/Release/v140/ or mingw/win32/
    }
    if on unix {
        try find libsodium from pkgconfig
        try find libsodium in env vars
        or build with ./configure && make
    }
} else {
    try find libsodium in env vars
      or

    if (crosscompiling to windows for msvc) {
        unpack prebuilt library e.g. msvc/x64/Release/v140/
    }
    if (crosscompiling to windows for gnu) {
        unpack prebuilt library e.g. mingw/win32/
    }
    if (crosscompiling to wasm32-wasi) {
        unpack prebuilt library e.g. wasm32/wasi/libsodium.a
    }

    if (crosscompiling on windows) {
        fail "cannot crosscompile on windows"
    }


    if (crosscompiling on unix to ios or mac or ...) {
        ./configure with CFLAGS && make
    }
}
```

Does it make sense to you?

@jean-airoldie you can fix this algo to provide version.h for windows (https://github.com/sodiumoxide/sodiumoxide/issues/370)
@stouset you can try to fix this algo to unify environment variables with those used in pkg-config (https://github.com/sodiumoxide/sodiumoxide/issues/359)

@dnaq what do you think?

jean-airoldie:
Seems like a `cc` build would be easier to maintain at this point.

kpp:
@jean-airoldie If only we could compile libsodium with crate cc...

```
~/rust_cross$ tree
.
├── build.rs
[...]

PR #376, opened 2019-09-18
Add init support for --target wasm32-wasi

kpp:
Closes: https://github.com/sodiumoxide/sodiumoxide/issues/347

Linked issues: https://github.com/jedisct1/libsodium/issues/872

PR #379, opened 2019-10-09
WIP: Try to build libsodium with cc only

kpp:
unpack libsodium to libsodium-sys/libsodium-1.0.18/ to build it

### DIFF
--- a/libsodium-sys/Cargo.toml
+++ b/libsodium-sys/Cargo.toml
@@ -15,16 +15,9 @@ version = "0.2.4"
 travis-ci = { repository = "sodiumoxide/sodiumoxide" }
 
 [build-dependencies]
-pkg-config = "0.3"
 tar = "0.4"
 libflate = "0.1"
-
-[target.'cfg(target_env = "msvc")'.build-dependencies]
-libc = { version = "0.2" , default-features = false }
-vcpkg = "0.2"
-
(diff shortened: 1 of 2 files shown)


kpcyrd:
This would drop support for dynamic linking, right? Can we keep pkg-config to check if libsodium is installed on the system and the version is >= our minimum version, and only fall back to static linking if that isn't the case?

kpp:
It's just a playaround. In https://github.com/sodiumoxide/sodiumoxide/pull/376 I discovered that we compile libsodiumoxide-rs in a wrong way which led to https://github.com/sodiumoxide/sodiumoxide/issues/377. So this example works for my target (linux x86-64) plus --target wasm32-wasi expect for 1 test with mprotect. I want to figure out whether it is possible to build libsodium on our own which can give the user a possibility of xcompilation and we won't need to keep blobs in our code anymore.

ryankurte:
ahh, something like this would be excellent, i had the same issue for the embedded / arm targets so have been patching libsodium-sys with [this](https://github.com/jedisct1/libsodium/compare/master...ryankurte:feature/cargo-compat).

should be easy to add a feature to choose between static or dynamic linking (pkgconfig / lib vs. build from source) if that'd get something like this landed?

kpp:
Seems like this PR won’t get landed

kpp:
Maybe someday in the future I will try it again.

PR #381, opened 2019-10-11
Fix critical Digest comparison bug

AbsurdlySuspicious:
`PartialEq` implementation for `generichash::Digest` has compared itself to itself

### DIFF
--- a/src/crypto/generichash/digest.rs
+++ b/src/crypto/generichash/digest.rs
@@ -26,7 +26,7 @@ impl PartialEq for Digest {
         if other.len != self.len {
             return false;
         }
-        memcmp(self.as_ref(), self.as_ref())
+        memcmp(self.as_ref(), other.as_ref())
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/crypto/generichash/digest.rs
```diff
@@ -26,7 +26,7 @@ impl PartialEq for Digest {
         if other.len != self.len {
```
kpp: BTW are you sure eq should be O(1) time?

kpp:
cc @RustSec

src/crypto/generichash/digest.rs
AbsurdlySuspicious: Same check is also done in the memcmp wrapper 
https://github.com/sodiumoxide/sodiumoxide/blob/master/src/utils.rs#L21

dnaq:
Thanks for finding this critical bug, and shame on us for not finding it a long time ago, preferably before merging. We need to think about how to handle this to get the community to switch to an updated version as soon as possible.

src/crypto/generichash/digest.rs
dnaq: Can you use the memcmp wrapper instead?

kpp: Then `if other.len != self.len {` in this eq operator is redundant.

kpp:
bors: r+

PR #388, opened 2019-11-08
Memzero-ing methods for Digest and generichash

AbsurdlySuspicious:
Memzero variants for consuming methods; for cases when blake2b is used for confidential data (key deriving, etc)


src/utils.rs
```diff
@@ -90,6 +91,14 @@ pub fn add_le(x: &mut [u8], y: &[u8]) -> Result<(), ()> {
     }
 }
 
+pub(crate) fn memzero_value<T>(mut v: T) {
```
adamreichold: Maybe this should indicate that it is actually dropping/forgetting value, like `memzero_and_forget_value`? (And maybe `debug_assert!(!needs_drop(...));`?)

adamreichold: Or `drop_memzero` for consistency with `Digest`?

src/crypto/generichash/mod.rs
```diff
@@ -101,6 +102,24 @@ impl State {
             Err(())
         }
     }
+
+    /// `finalize` finalizes the state and returns the digest value. `finalize` consumes the
+    /// `State` and securely erases it using `utils::memzero`. Use this function for confidential data.
+    pub fn finalize_memzero(mut self) -> Result<Digest, ()> {
```
adamreichold: I guess this could also be layered on top of `finalize` to avoid repetition.

src/utils.rs
adamreichold: Actually, one could avoid changing the API of `Digest` and `State` by just providing a public `forget_memzero` function that could just be used like `forget_memzero(digest);` which is not more complicated than `digest.drop_memzero();` and aligns nicely with the standard libraries `mem::forget` and `mem::drop` functions.

dnaq: This function is unsafe. Being able to zero out any value will cause undefined behavior. Also it won’t work for any type that keeps its values anywhere else than in the struct, causing values to stay in memory, memory leaks and so on.

adamreichold: > This function is unsafe. Being able to zero out any value will cause undefined behavior. 

I though so too at first, but having taken ownership of the value and `forget`ting it before zeroing out the memory should actually be safe as nothing is touching the memory anymore.

> Also it won’t work for any type that keeps its values anywhere else than in the struct, causing values to stay in memory, memory leaks and so on.

Indeed, but failing to clear everything and leaking allocations is not a memory safety issue. (`forget` itself is not unsafe either). But a `T: Copy` bound would probably be a good thing if it is intended to improve security.

One other issue issue is taking ownership could imply `memcpy` behind the scenes, but AFAIK this affects the other zero-on-drop types as well.

adamreichold: The leaking is actually not strictly necessary either as `ptr::drop_in_place` could be used instead of `mem::forget`.

dnaq: Leaking isn’t unsafe, it’s just bad. Overwriting memory with bit patterns the type can’t contain is undefined behavior, no matter if the value is dropped afterwards or not, and no matter if the type is copy or not. See the whole raison d’etre for MaybeUninit.

adamreichold: > Overwriting memory with bit patterns the type can’t contain is undefined behavior, no matter if the value is dropped afterwards or not, and no matter if the type is copy or not. See the whole raison d’etre for MaybeUninit.

Interpreting an arbitrary a bit pattern as `T`, e.g. `bool`, is UB. But zeroing memory at an arbitrary address if there is no more `T` is not. `MaybeUninit` makes it easier to work with uninitialized memory and avoid the pitfalls of `mem::uninitialized`, but there is nothing preventing one from doing all of the same using only raw pointers.

In this case, this is exactly like deallocting a heap-allocated value, e.g. [`impl Drop for Rc`](https://doc.rust-lang.org/src/alloc/rc.rs.html#1065-1110) (`Box` is special at it is currently impemented in the compiler), i.e. the value is destroy using `ptr::drop_in_place` and then it is deallocated using `Global.dealloc`. Nothing prevents us from zero'ing out the memory in between these steps as there is no more `T` the invariants of which we need to uphold.

adamreichold: (Note that there is no `&mut T` or even `*mut T` in play after `drop`ping, only `*mut u8` resp. `&mut [u8]`.)

adamreichold: I think `MaybeUnint` would actually make this nicer to write, e.g.
```
fn drop_memzero<T: Copy>(mut v: T) {
  let v = mem::MaybeUninit::new(v);
  unsafe {
        ptr::drop_in_place(v.as_mut_ptr());
        ffi::sodium_memzero(v.as_mut_ptr() as _, mem::size_of::<T>());
  }
}
```

AbsurdlySuspicious: `forget_memzero(state)` isn't possible anyway since `finalize` consumes the state

adamreichold: > forget_memzero(state) isn't possible anyway since finalize consumes the state

Right, I did not see that. But then, the actual finalize logic should probably be factored out into a `&mut self` method to avoid repeating it. The question whether `drop_memzero` resp. `memzero_value` are safe still needs to answered tough.

AbsurdlySuspicious: State is invalid after finalizing. Original libsodium documentation clearly states that state can't be used again after finalizing, so it needs to be consumed.

adamreichold: > Original libsodium documentation clearly states that state struct can't be used again after finalizing, so it needs to be consumed.

Yes, and the public API should enforce this. That does not mean that an internal helper method could not take `&mut self` to finalize, it should just not be public.

AbsurdlySuspicious: Yeah, it makes sense. I'll push somethig on that matter soon

AbsurdlySuspicious:
+ Deduplicated finalize code
+ More idiomatic approach for `memzero_value` using MaybeUninit

src/crypto/generichash/digest.rs
```diff
@@ -14,6 +15,13 @@ pub struct Digest {
     pub(super) data: [u8; crypto_generichash_BYTES_MAX as usize],
 }
 
+impl Digest {
+    /// Securely erase `Digest` using `utils::memzero`
+    pub fn drop_memzero(mut self) {
```
adamreichold: I'd still this would be better served by making `memzero_value` a public API with a `Copy` bound.

I'd also say the doc comment wording is misleading as this is not "secure", it is merely best effort as `Digest` is as movable as any Rust type and could have been `memcpy`ied multiple times so that this will only zero out the last copy which will actually just be the `self` argument pass to this method without further compiler optimizations.

src/utils.rs
```diff
@@ -90,6 +91,14 @@ pub fn add_le(x: &mut [u8], y: &[u8]) -> Result<(), ()> {
     }
 }
 
+pub(crate) fn memzero_value<T>(v: T) {
```
adamreichold: This should have `T: Copy` bound or call `ptr::drop_in_place`. Considering the intended use case, the `Copy` bound seems more appropriate.

AbsurdlySuspicious: Whats the purpose in memzero-ing copied data?

src/crypto/generichash/digest.rs
AbsurdlySuspicious: Same does all the key wrappers and still they have a drop implementation with zeroing. Also, the decision to publish that function or not isn't up to me.

src/utils.rs
adamreichold: `T: Copy` does not mean that the data is copied, it means that it does not indirectly own any resources via e.g. pointer direction so that `Clone`ing it means to byte-for-byte copy it. If zeroing it out is to help with security, we need to avoid data types with indirections or follow all of the indirections to really zero out the relevant memory.

src/crypto/generichash/digest.rs
adamreichold: > Same does all the key wrappers and still they have a drop implementation with zeroing. 

I am aware of that but they do not say anything about "securely" erasing things in their doc comments and I think we should avoid that here as well.

src/utils.rs
dnaq: But in practice it means that the struct will be memcpy’d into this function and then the copy will be securely erased. The compiler is free to give a reference, but in many cases a memcpy will happen which makes the whole function meaningless. You can verify this for many cases by looking at the generated assembly.

adamreichold: > But in practice it means that the struct will be memcpy’d into this function and then the copy will be securely erased. 

This is true and was mentioned in both other threads here. But it does also affect the existing memzero-on-drop types as they can be moved - meaning `memcpy`ied around - just as well and `drop` will only zero out the last copy. Taking `self` by value only adds one more place where this problem manifests, but the problem itself is already present in the other wrapper types.

If we really want to fix this, then all sensitive data types have to be put into stable locations, e.g. using `Box` or `StableDeref` in general, and zeroing-on-drop must only work for those types and we must wrap them so user code will not accidentally copy/move out of them.

dnaq:
The current APIs only do a best effort cleaning, just as you say, but they
allow the user to make the choice if needed (and to be honest, most users
don’t need to make that choice, since if someone can dump their running
memory they have other more urgent problems), however this function will
always memcpy, and will thus be meaningless.

src/crypto/generichash/digest.rs
AbsurdlySuspicious: Then please propose doc-comment that in your opinion wouldn't be confusing

src/utils.rs
adamreichold: > but they allow the user to make the choice if needed

I do not understand that. Which choice exactly do you mean? The compiler is free to move values around without user code explicitly requesting it and it can avoid to do so even if user code contains a move.

> however this function will always memcpy, and will thus be meaningless.

Not always, but yes the likelihood is high especially for small types. If we want to give something reliable here, we should restrict ourselves to something like `Box<T> where T: Copy` to get a stable pointer to zero-out before deallocation.

adamreichold: > (and to be honest, most users don’t need to make that choice, since if someone can dump their running memory they have other more urgent problems)

I somewhat agree, but following this means that this API is not really worth the effort in the first place. Those very few users you need it can just unsafely call `utils::memzero` directly while having the highest chance to succeed using e.g. stable allocations.

adamreichold: (I am not sure about the ergonomics of this, but `Pin<&mut T>` might also be useful choice.)

AbsurdlySuspicious: Btw why so centered on not changing the api? Changes proposed in this PR is not destructive and allowed by rust semver convention (for next minor version)

dnaq:
I would be more ok with just adding the Drop implementation to Digest, like
we do with the other secret types. If it can contain secrets then we try to
erase it.

src/utils.rs
AbsurdlySuspicious: > I would be more ok with just adding the Drop implementation to Digest, like
we do with the other secret types. If it can contain secrets then we try to
erase it.

But generichash can also be used for non-confidential data and then forcefully erasing it might be a performance hit, even if minor

adamreichold: > Btw why so centered on not changing the api? 

This is not my motivation. I'd just rather provide a general mechanism than do this again and again for every type. Especially since it really is optional for a digest. Also for `State`, there is still the problem of making the compiler take ownership but elide copies as much as possible.

> I would be more ok with just adding the Drop implementation to Digest, like
we do with the other secret types. If it can contain secrets then we try to
erase it.

This could imply a significant overhead as there are usually many more instances of digests than of e.g. keys. It would also imply that this should happen for SHA-256/512 as well. Making everyone pay that overhead for a best-effort mechanism is a questionable trade-off IMHO. Especially since the `Memzero` wrapper struct would it trivial to add this would modifying the `Digest` type at all.

Even for `State` something like
```
impl Memzero<T> {
  pub fn consume<F, R>(self, f: F) -> R where F: FnOnce(T) -> R {
    f(unsafe { ptr::read(self.0.as_ptr()) })
  }
}
```
would allow unmodified usage like `let digest = state.consume(State::finalize);` with less
 copying than the current proposal.

AbsurdlySuspicious: > every type

I can't cone up with any type other than `Digest` and `generichash::State` for which this behavior would desired to be optional. All other types known to contain private data have automatically implemented `Drop`

AbsurdlySuspicious: If you want to provide a way for users to erase their own structures, I think you should open separate PR as this is not really related to Digest.

adamreichold: > All other types known to contain private data have automatically implemented Drop

What is different about `sha256::Digest` or `sha512::Digest` compared to `generichash::Digest`?

> If you want to provide a way for users to erase their own structures, I think you should open separate PR as this is not really related to Digest.

I think providing an API to zero out arbitrary `Copy` types is the most effective API to achieve this, especially since it implies that no modifications to `Digest` or `State` are necessary. I you want I can propose a separate MR for this, but I do think it would make this one unnecessary which is way I suggested modifying this one.

adamreichold: #394

AbsurdlySuspicious: > What is different

Nothing. They still are digests and just need the same api as blake2b digest. I missed them myself when implementing this, but I don't think it's worth to dance with state consuming instead of just extending digest api.

adamreichold: > but I don't think it's worth to dance with state consuming instead of just extending digest api.

I had not though of that yet, but there is actually important ergonomic benefit to wrapper types: One cannot forget to call the zeroing `Drop´ implementation whereas it is quite possible to write a function that has some early returns, e.g. for error handling, that fail to call `drop_memzero`, `finalize_memzero` and friends. One would then probably put these calls into a scope guard and end up with something resembling the wrapper types in a more roundabout fashion.

kpp:
Hello. Sorry for the delay. We have changed our CI a little bit.

@AbsurdlySuspicious would you please rebase onto the latest master.

@adamreichold  would you please resolve threads which seem to be resolved?

adamreichold:
> would you please resolve threads which seem to be resolved?

Maybe I am missing something, but I don't think I can resolve threads in a PR that I did not create without project access.

AbsurdlySuspicious:
By the way, I'll also summarize options about this PR:

- Discard this PR completely
- Keep new functions as-is and clarify docs about this being best-effort zeroing
- Add zeroing to Drop trait as with secret key structs (may have performance penalty since hashes are not always secret data)
- Try to somehow rule out copying with Pin (?)
- Maybe make something like SecretDigest in case manual dropping functions are really not desirable

adamreichold:
> Maybe make something like SecretDigest in case manual dropping functions are really not desirable

That could for example be `Memzero<Digest>` from #394

kpp:
@AbsurdlySuspicious @adamreichold I see you worked on almost the same issue in separate PRs. Would you please come up to a conclusion which PR we should continue working on?

adamreichold:
> @AbsurdlySuspicious @adamreichold I see you worked on almost the same issue in separate PRs. Would you please come up to a conclusion which PR we should continue working on?

If read the [last comment](https://github.com/sodiumoxide/sodiumoxide/pull/388#issuecomment-556045150) by @dnaq correctly, then modifying this PR to

> Add zeroing to Drop trait as with secret key structs (may have performance penalty since hashes are not always secret data)

was his favorite.

I think #394 can provide the same results without the performance penalty, but as far as I understand @AbsurdlySuspicious and me still disagree on this and we need a maintainer's decision to resolve this. (Please note that #394 was result of the dicussion here, so these are not parallel efforts but rather a single discussion spanning two PR.)

dnaq:
I think we might go with zero on drop for now, best effort. I think #394 is
a good idea, but I’m not a fan of the exposed api (consuming the value with
a closure), I think a better api would be a type that Boxes memory and
gives references to it, but that won’t work on no alloc environments.

dnaq:

PR #390, opened 2019-11-19
Expose libsodium's hexadecimal encoding/decoding helper functions

adamreichold:
This allows dependent crates to avoid an additional dependency for this and it avoids the development dependency on `hex` for this crate.

### DIFF
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -1,0 +1,83 @@
+//! Libsodium hexadecimal encoding/decoding helper functions
+use ffi;
+#[cfg(not(feature = "std"))]
+use prelude::*;
+use std::ptr;
+
+/// Encodes byte sequence into a hexadecimal string.
+///
+/// # Panics
+///
+/// Panics if `2 * bin.len() + 1` overflows.
+pub fn encode<T: AsRef<[u8]>>(bin: T) -> String {
+    let bin = bin.as_ref();
+    let mut hex = vec![0; encoded_len(bin.len()).unwrap()];
+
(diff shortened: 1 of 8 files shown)


src/hex.rs
```diff
@@ -0,0 +1,83 @@
...
+#[cfg(not(feature = "std"))]
+use prelude::*;
+use std::ptr;
+
+/// Encodes byte sequence into a hexadecimal string.
+///
+/// # Panics
+///
+/// Panics if `2 * bin.len() + 1` overflows.
+pub fn encode<T: AsRef<[u8]>>(bin: T) -> String {
```
kpp: Why decode returns `Result::Err` whereas encode panics on overflow?

adamreichold: I am not really invested in that and will happily make `encode` return `Result<String, ()>` if you prefer.

The reasoning that led me here was:
* Overflowing `usize` is highly unlikely as it is hard to even have such large allocations and hence a good candidate for a panic.
* As this is the only failure more of `encode`, it does not need to return `Result`.
* For `decode`, there is the failure mode of invalid input which I have to handle as a `Result` because it is not under our control.
* If `decode` has to return anyway, I can just as well turn the panic into an error return.

Another way to go would be to make `decode` panic in the overflow case as well. But as indicated above, I think this is a judgement call and will be happy to do it in any of these ways.

adamreichold:
Sorry for resetting the review state: I rebased this to resolve a trivial merge conflict in `src/lib.rs` and fix two typos in the just merged `base64` module.

kpp:
bors: r+

PR #391, opened 2019-11-19
Expose libsodium's Base64 encoding/decoding helper functions

adamreichold:
This allows dependent crates to avoid an additional dependency for this.


kpp:
bors: r+

PR #393, opened 2019-11-20
Ship libsodium as a Git submodule

### LINKED ISSUE #392: Alternatives to build dependency on tar and libflate
For me, a clean build `libsodium-sys` build time is currently dominated by building `tar` and `libflate` and their dependencies instead of build libsodium itself. These crates are used to unpack the source archive that is shipped with this crate when libsodium is built from source (the default).

Personally, I would be glad if these two build dependencies could be avoided and see three possible approaches:
* Do not ship a compressed tarball inside the crate (which already is a compressed archive) but just commit the libsodium sources directly to e.g. `libsodium-sys/libsodium`.
* The above but as a Git submodule pointing at tags in `https://github.com/jedisct1/libsodium.git` whichs seems quite common for many `*-sys` crates.
* Shell out to `tar` and `gzip` for decompression as we only need this when building from source where we assume a full build chain including autotools anyway so that these two do not seem to make things worse.

adamreichold:
Avoid dev dependencies on tar and libflate by including libsodium as a Git submodule.

Closes #392


adamreichold:
> It is okay for cargo to build multiple builds concurrently: debug, release, rls. So you will have to have a copy of libsodium per each build.

But this is exactly as it currently is, i.e. we untar the archive for each build. The only thing I changed is that I copy an already unpacked directory tree instead of untar'ing it.

> Also you will have to git clone atomically because a build can be canceled during git clone.

Cloning is not part of the build process. Either it is done when this repository here is initially cloned or Cargo will package the source when it ships the crate to crates.io as a tarball, e.g. c.f. https://github.com/alexcrichton/xz2-rs/tree/master/lzma-sys and https://docs.rs/crate/lzma-sys/0.1.15/source/

> Also your method requires preinstalled robocopy or git. 

It does not require Git as Cargo will package the source when shipping the crate. Robocopy is installed by default since Windows Vista but I can certainly use xcopy as well, it is just that xcopy is deprecated since Windows 10.

adamreichold:
AFAIK, Cargo will even clone recursively when one uses a Git dependency directly in `Cargo.toml`, so even in that case one does not need to do anything special or install Git as Cargo contains libgit2 built from a submodule, c.f. https://github.com/rust-lang/git2-rs/tree/master/libgit2-sys

adamreichold:
AppVeyor builds seem to fail randomly as connections to `github.com:443` time out...

kpp:
I see. Thanks.

adamreichold:
Sorry for the noise but could someone retry that one failed (when downloading  7-Zip) AppVeyor job? I seem to have insufficient privileges to reach that part of the UI...

kpp:
Neither can I. Please do `git commit --amend && git push -f`

kpp:
error: failed to run custom build command for `libsodium-sys v0.2.5 (/home/humbug/sodiumoxide/libsodium-sys)`

Caused by:
  process didn't exit successfully: `/home/humbug/sodiumoxide/target/debug/build/libsodium-sys-b1901ced0a30985f/build-script-build` (exit code: 101)
--- stdout

--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:1165:5
stack backtrace:

  15: core::result::Result<T,E>::unwrap
             at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/result.rs:933
  16: build_script_build::make_libsodium
             at libsodium-sys/build.rs:249
  17: build_script_build::build_libsodium
             at libsodium-sys/build.rs:425
  18: build_script_build::main
             at libsodium-sys/build.rs:62

adamreichold:
> I got this error on your branch:

Will try to reproduce and fix.

adamreichold:
> > I got this error on your branch:
> 
> Will try to reproduce and fix.

But actually, I can't at least running on Linux. I have tried both `cargo clean` and `rm -rf target` and both work for me. Can you specify which platform you are using? Could list the contents of `target/debug/build/libsodium-sys-*/out` for your build?

(I had the same error for an intermediate version on which I force-pushed. Are you sure you are on e09f0f2?)

kpp:
I see. I had to `~/sodiumoxide$ git submodule update --init`. This should be included into readme for devs.

adamreichold:
>  I had to ~/sodiumoxide$ git submodule update --init. This should be included into readme for devs.

Sorry for the confusion. Will amend the README...

adamreichold:
> > I had to ~/sodiumoxide$ git submodule update --init. This should be included into readme for devs.
> 
> Sorry for the confusion. Will amend the README...

I have an extended version, please have a look to check if you think this is enough. (I chose and set of commands that should work for all Git versions even tough nowadays one would probably just use `git clone --recurse-submodules`.)

kpp:
Unfortunately a local dep `sodiumoxide = {path = "../sodiumoxide"}` w/o `git submodule update --init` will fail too

adamreichold:
> Unfortunately a local dep `sodiumoxide = {path = "../sodiumoxide"}` w/o `git submodule update --init` will fail too

But I think this is very much expected as a path dependencies means you mean complete but hence manual control of the target crate's source, i.e. you are the one who did the `git clone`ing, in contrast to a Git dependency where Cargo does itself and includes submodules.

kpp:
Sounds like an argument.

Stll I want a human error msg for error `https://github.com/sodiumoxide/sodiumoxide/pull/393#issuecomment-557789422`

adamreichold:
> Stll I want a human error msg for error [#393 (comment)](https://github.com/sodiumoxide/sodiumoxide/pull/393#issuecomment-557789422)

Amended this to suggest checking the submodule.

kpp:
> thread 'main' panicked at 'Failed to find configure script! Did you clone the submodule at `libsodium-sys/libsodium`?

Nice! Thanks

kpp:
@sodiumoxide/owners please review

README.md
```diff
@@ -156,7 +163,7 @@ linker = "arm-buildroot-linux-musleabihf-gcc"
 2. Dockerfile:
 
 ```
-FROM rust:1.30.1
+FROM rust:1.36.0
```
Dylan-DPC: Is there a reason for the version bump? I would prefer it in another pr

adamreichold: The README states the minimum Rust version for this crate to be 1.36.0 and when I tested cross building for another MR and I ran into this and hence I though I boy scout it here. But you are right insofar it does not really belong here and I'll move it into a separate MR.

Dylan-DPC: Ah yes. It is fine in this case

adamreichold: > But you are right insofar it does not really belong here and I'll move it into a separate MR.

Done.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libsodium-sys/libsodium"]
+	path = libsodium-sys/libsodium
+	url = https://github.com/jedisct1/libsodium.git
```
kpp: Is it v1.0.18 or master?

adamreichold: [The submodule](https://github.com/sodiumoxide/sodiumoxide/pull/393/files#diff-56e707f409b4a26a091f2f02232abafe) is at [940ef4](https://github.com/jedisct1/libsodium/tree/940ef42797baa0278df6b7fd9e67c7590f87744b) which corresponds to [1.0.18-RELEASE](https://github.com/jedisct1/libsodium/releases/tag/1.0.18-RELEASE), so this should be v1.0.18. (There is another tag called [1.0.18](https://github.com/jedisct1/libsodium/releases/tag/1.0.18) from the say date, but it did not advertise a tarball.)

kpp: Thanks

adamreichold:
@kpp Anything I can do to help with the decision? I'd be happy to add documentation, try another variation of this, etc.

kpp:
> Anything I can do to help with the decision?

I like the idea. The less deps the better. I want one more + to merge this.

adamreichold:
> > Anything I can do to help with the decision?
> 
> I like the idea. The less deps the better. I want one more + to merge this.

I understand and will be more patient. Thanks!

dnaq:
I think it’s a very good change. Maybe just bump the libsodium version to the latest one in the stable branch.

kpp:
> Maybe just bump the libsodium version to the latest one in the stable branch.

Right now we compare pkg_config's version of libsodium with "1.0.18". Will there be "stable" in the version of pkg_config too?

adamreichold:
> > Maybe just bump the libsodium version to the latest one in the stable branch.
> 
> Right now we compare pkg_config's version of libsodium with "1.0.18". Will there be "stable" in the version of pkg_config too?

Personally, I don't mind either way as I am not familiar with the handling of libsodium's `stable` branch. But I also think the issue is separate from how we package it (using a submodule just makes it easier to change this), so I'd suggest to keep this minimal, i.e. just changing how to package but not what, but I'd be glad to prepare a follow-up pull request to bump the submodule to e.g. [1d4017df58d87822bca16690fcc0e1a5af00a3a5](https://github.com/jedisct1/libsodium/commit/1d4017df58d87822bca16690fcc0e1a5af00a3a5).

kpp:
bors: r+

adamreichold:
@kpp Not sure if someone stopped bors because he or she needs more time to review or if bors just cancelled because of the missing GitHub review here?

kpp:
I guess you have to update the branch on the latest master

adamreichold:
> I guess you have to update the branch on the latest master

Rebased on the current `master`.

PR #394, opened 2019-11-20
Add wrapper to zero out arbitrary types on drop

adamreichold:
This allows to zero out types like `generichash::{State, Digest}` which are sometimes performance critical, but contain sensitive information in other cases so that always zeroing out on drop could be considered prohibitively expensive.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,11 +1,64 @@
 //! Libsodium utility functions
 use ffi;
+use std::{
+    mem,
+    ops::{Deref, DerefMut},
+    ptr,
+};
 
 /// `memzero()` tries to effectively zero out the data in `x` even if
 /// optimizations are being applied to the code.
 pub fn memzero(x: &mut [u8]) {
     unsafe {
         ffi::sodium_memzero(x.as_mut_ptr() as *mut _, x.len());
+    }
+}
+
+/// Wrapper type that tries to effectively zero out the contained data
+/// during drop. This should not be used with types containing indirections
+/// like heap-allocated memory as it cannot be effective in these cases.
+pub struct Memzero<T>(mem::MaybeUninit<T>);
+
+impl<T> Memzero<T> {
+    /// Create a wrapped value
+    pub fn new(val: T) -> Self {
+        Self(mem::MaybeUninit::new(val))
+    }
+
+    /// Consume a wrapped value using the given closure
+    pub fn consume<F, R>(self, f: F) -> R
+    where
+        F: FnOnce(T) -> R,
+    {
+        // SAFETY: We will only deinitialize `self.0` inside `drop`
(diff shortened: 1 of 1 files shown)


AbsurdlySuspicious:
Using Copy as any kind of guarantee is really a misuse since:
- rustc will happily memcpy any struct marked as Copy when passed anywhere, rendering useless not only zeroing, but making a whole bunch of other problems
- It doesn't really guarantee anything. It relies solely on the fact that `&mut T` is not `Copy`. However, both const and mut raw pointer types are

If this kind of guarantee is really needed (which I doubt), it can be made with custom (unsafe) marker trait. While limiting use of this wrapper to sodiumoxide structs and other explicitly marked types, it will provide mentioned guarantees (for real this time) without other drawbacks

adamreichold:
> rustc will happily memcpy any struct marked as Copy when passed anywhere, rendering useless not only zeroing, but also making a whole bunch of other problems

Passing `&T` will not `memcpy` whether `T: Copy` or not. But I do agree that `Copy` makes it easier to accidentally copy. 

> It doesn't really guarantee anything. It relies solely on the fact that &mut T is not Copy. However, both const and mut raw pointer types are

I can not really follow this. Which guarantee do you think this is relying on and how does `&mut T` enter the picture? 

> If this kind of guarantee is really needed (which I doubt), it can be made with custom (unsafe) marker trait. While limiting use of this wrapper to sodiumoxide structs and other explicitly marked types, it will provide mentioned guarantees (for real this time) without other drawbacks

`Copy` is used to prevent types that make use of indirection as this would defeat the purpose and leak indirectly owned data, e.g. there can be no `Drop` implementations for `Copy` types.

The trait bound can certainly be removed entirely and no unsafe marker trait is necessary, as one can certainly zero out any type inside `MaybeUninit` thereby leaking any indirectly owned data. It is just a question of usability to help users of `Memzero` to avoid applying it ineffectively. So a - normal, not unsafe - trait could be used to achieve marking types of which can be effective, but it seems a questionable extra step for a best-effort mechanism IMHO.

So I'd suggest just removing the bound completely and documenting that this will leak its insides and will not work on types using indirection?

adamreichold:
Of course, I could call `drop_in_place` on the innards of `Memzero` before zeroing to avoid leaking if someone accidentally used this as `Memzero<Box<T>>`, but I think don't think we should cater to such misuse. Another option would be to have a runtime check based on `needs_drop` in the constructor and panic if there is a non-trivial `Drop` impl as this can never be a correct usage of this type.

adamreichold:
Thinking about this again, there could be `Drop` implementation of types which could be `Copy` that interact with some global/external state to perform some clean-up etc. So I guess I could just add the `drop_in_place` call and document how this should be used?

PR #373, opened 2019-09-09
remove ill-advised gen_nonce functions

GallagherCommaJack:
solves #91 without breaking tests by making functions `pub(crate)` by default, creating `pub` variants where appropriate.

I also changed the default `aead` construction to bring it in line with libsodium's recommendations instead of favoring the less-secure primitive.

PR #395, opened 2019-11-23
Fix unused warnings as  is always defined but not always exported

adamreichold:
Warnings like https://travis-ci.org/sodiumoxide/sodiumoxide/jobs/615562099#L255

### DIFF
--- a/src/crypto/aead/aead_macros.rs
+++ b/src/crypto/aead/aead_macros.rs
@@ -55,6 +55,7 @@ pub fn gen_key() -> Key {
 /// THREAD SAFETY: `gen_nonce_internal()` is thread-safe provided that you have
 /// called `sodiumoxide::init()` once before using any other function
 /// from sodiumoxide.
+#[allow(unused)]
 fn gen_nonce_internal() -> Nonce {
     let mut n = Nonce([0u8; NONCEBYTES]);
     randombytes_into(&mut n.0);
--- a/src/crypto/aead/aead_macros.rs
+++ b/src/crypto/aead/aead_macros.rs
@@ -50,18 +50,6 @@ pub fn gen_key() -> Key {
     k
 }
 
-/// `gen_nonce_internal` randomly generates a nonce
-///
-/// THREAD SAFETY: `gen_nonce_internal()` is thread-safe provided that you have
-/// called `sodiumoxide::init()` once before using any other function
-/// from sodiumoxide.
-#[allow(unused)]
-fn gen_nonce_internal() -> Nonce {
-    let mut n = Nonce([0u8; NONCEBYTES]);
-    randombytes_into(&mut n.0);
-    n
-}
-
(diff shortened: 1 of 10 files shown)


kpp:
How about unit tests? The report showed 0.3% coverage decreasing https://github.com/sodiumoxide/sodiumoxide/pull/373#issuecomment-529317478

adamreichold:
> Maybe it would be better than hiding warnings?

Unit tests won't make these errors go away: Tests are compiled using `#[cfg(test)]` and hence a normal build of this library, e.g. as a dependency for just via `cargo clippy`, will still produce them.

This is dead code after all as non-public functions can't be used by anything outside the crate itself and adding tests does not change that. Or put differently, why test something that nobody using this library can actually use.

I think the cleaner way would be to not define the the internal functions at all. Maybe actually provide a generic method like
```
fn gen_random_nonce<N: Default + AsMut<[u8]>>() -> N {
  let mut n = N::default();
  randombytes_into(n.as_mut());
  n
}
```
instead of generating via a macro? But this would imply implementing `AsMut<[u8]>` for `new_type!{nonce ..}`...

> How about unit tests? The report showed 0.3% coverage decreasing [#373 (comment)](https://github.com/sodiumoxide/sodiumoxide/pull/373#issuecomment-529317478)

I can add more unit tests if you like but I have to admit I feel that they should have part of the original MR. Can you be specific which unit tests you'd like to see?

kpp:
> I feel that they should have part of the original MR

Good idea

adamreichold:
> Maybe actually provide a generic method like

I added another commit here that implements the generic method (c.f. the commit message for the details) and avoid defining any unused functions or extending the public interface of the nonce types. I feel like this is a nicer solution overall, but it admittedly is also more involved than the status quo.

kpp:
Looks like the PR did not affect coverage

adamreichold:
> Looks like the PR did not affect coverage

Different from what I implied earlier, the loss of coverage in #373 was not because of the unused functions but due to the forwarding functions `gen_nonce` which still seem to be in need of additional unit tests as the unit tests generated by the macros do not cover them anymore. But then these unit tests would be copies of the ones created by the macros...

adamreichold:
> But then these unit tests would be copies of the ones created by the macros...

Actually, I'll try to add something that asserts on the nonce length which should actually be somewhat sensible and should fix the coverage.

adamreichold:
> > But then these unit tests would be copies of the ones created by the macros...
> 
> Actually, I'll try to add something that asserts on the nonce length which should actually be somewhat sensible and should fix the coverage.

Did that. Also factored out the random nonce generation from the `box_` and `secretbox` modules.

kpp:
Ready to merge?

adamreichold:
> Ready to merge?

Yes, definitely. I was just adding more cosmetics. Sorry for the large number of force-pushes. I just try to get review as early as possible and then always find more nits to fix.

kpp:
bors: r+

PR #397, opened 2019-11-23
RFC: Deny Clippy warnings in CI jobs to catch more issue automatically

adamreichold:
At $DAYJOB, we run the CI with `cargo clippy -- --deny warnings` to essentially avoid having them in the code base but not making the edit-build-test cycle harder as with a `lib.rs`-based configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
     before_script:
       - rustup component add clippy
     script:
-      - cargo clippy --all --tests
+      - cargo clippy --all --lib --tests -- --deny warnings
   - name: coverage
     os: linux
     rust: stable


kpp:
bors: r+

adamreichold:
@kpp I just cancelled bors because I noticed that this should fail until #395 is merged. But it did not because it ran Clippy only in `#[cfg(test)]` but not without it. Ideally one would use `--all-targets` but since the benchmarks in this crate will only work on Nightly, I opted for `--lib --tests` for now to get more coverage of the code base. But of course, this implies that this cannot be merged before #396 or at least something similar.

kpp:
Also would be nice to have -Dwarnings

adamreichold:
Rebased on `master` so that it should have chance to pass the CI now.

> Also would be nice to have -Dwarnings

I am not sure I follow: Isn't `-Dwarnings` short hand notation for `--deny warnings` as added here? Or do you mean for other jobs? Or locally when not built by the CI?

kpp:
I was talking about https://github.com/async-rs/async-std/blob/50cefce803c88f3cd321b3bc181427b2bc4b30ce/.github/workflows/ci.yml#L11 which denies warnings even in normal builds

adamreichold:
> s talking about https://github.com/async-rs/async-std/blob/50cefce803c88f3cd321b3bc181427b2bc4b30ce/.github/workflows/ci.yml#L11 which denies warnings even in normal builds

Passing `--deny warnings` to Clippy will also deny any warnings emitted by rustc itself. So passing that flag in other builds will only imply that more of the Travis jobs fail, but the Clippy job alone is sufficient to catch all the errors.

kpp:
Nice

kpp:
bors: r+

PR #398, opened 2019-11-23
Expose libsodium's single-chunk generic hash routine

adamreichold:
Also exposes that libsodium provides a recommended minimum hash size (currently 256 bits) so that the user can be freed from having to come up with a reasonable hash size in many cases.

The API change of `State::new` is of course breaking which could be avoided in various ways:
* `State` just requires the user to supply a hash size and `hash` has a different API.
* A separate `State::rec_out_len` constructor is provided that forwards to an internal generic constructor.
* Expose the recommended digest length as a constant `DIGEST_REC` that the user must pass instead of `None`.

Personally, I opted against that as the current 0.2.x crate version suggests that the API is still evolving and hence a clean API seems preferable to a compatible one.

### DIFF
--- a/src/crypto/generichash/mod.rs
+++ b/src/crypto/generichash/mod.rs
@@ -1,9 +1,10 @@
 //! `GenericHash`.
 //!
 use ffi::{
-    crypto_generichash_BYTES_MAX, crypto_generichash_BYTES_MIN, crypto_generichash_KEYBYTES_MAX,
-    crypto_generichash_KEYBYTES_MIN, crypto_generichash_final, crypto_generichash_init,
-    crypto_generichash_state, crypto_generichash_update,
+    crypto_generichash, crypto_generichash_BYTES, crypto_generichash_BYTES_MAX,
+    crypto_generichash_BYTES_MIN, crypto_generichash_KEYBYTES_MAX, crypto_generichash_KEYBYTES_MIN,
+    crypto_generichash_final, crypto_generichash_init, crypto_generichash_state,
+    crypto_generichash_update,
 };
 
 use libc::c_ulonglong;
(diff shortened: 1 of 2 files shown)


kpp:
@adamreichold ping. Would you please resolve conflicts?

adamreichold:
> @adamreichold ping. Would you please resolve conflicts?

Done. I held off on it so far as the issue

> The API change of State::new is of course breaking which could be avoided in various ways:

mentioned in the cover letter still needs a resolution by a maintainer.

src/crypto/generichash/mod.rs
```diff
@@ -103,6 +90,61 @@ impl State {
...
+///
+/// `out_len` specifies the resulting hash size.
+/// Only values in the interval [`DIGEST_MIN`, `DIGEST_MAX`] are allowed.
+/// If omitted, a minimum recommended output size is used which makes it
+/// practically impossible for two messages to produce the same fingerprint.
+///
+/// `key` is an optional parameter, which when given,
+/// a custom key can be used for the computation of the hash.
+/// The size of the key must be in the interval [`KEY_MIN`, `KEY_MAX`].
+pub fn hash(out_len: Option<usize>, key: Option<&[u8]>, data: &[u8]) -> Result<Digest, ()> {
```
kpp, line 103: Why not put `data` in the first place?

kpp (dismissed):

kpp:
> The API change of State::new is of course breaking which could be avoided in various ways:

That's fine because we already broke API.

src/crypto/generichash/mod.rs
adamreichold: No particular reason. Will change.

kpp (approved):

kpp:
bors: r+

PR #400, opened 2019-12-04
Bump the libsodium submodule to the current stable branch.

adamreichold:
libsodium [recommends](https://download.libsodium.org/doc/installation#stable-branch) using its stable branch when building from source as it contains critical fixes since the last release.

We should probably decide on
- [x] How to handle the binaries we ship for MinGW and MSVC as they are from the tagged release: We'll ship binaries from the [stable-branch-based snapshot builds](https://download.libsodium.org/libsodium/releases/).
- [x] How to handle the version check we building against an external libsodium version.
before merging this: The version check is unchanged as the stable branch will always be compatible to the latest tagged release.


adamreichold:
Personally, I think keeping `pkg-config` and the binaries at 1.0.18 would be alright as the `stable` branch is supposed to contain only critical fixes and not API changes. We can bump the version and ship new binaries when a new release is made as we need to update our checkout of the stable branch as well.

adamreichold:
> Personally, I think keeping `pkg-config` and the binaries at 1.0.18 would be alright as the `stable` branch is supposed to contain only critical fixes and not API changes. We can bump the version and ship new binaries when a new release is made as we need to update our checkout of the stable branch as well.

I just noticed prebuilt Windows binaries for snapshots are published at https://download.libsodium.org/libsodium/releases/ as well, so we could update those whenever we update the submodule as well. I guess this becomes mainly a matter of documentation which stable versions the various platforms get?

kpp:
Try to build the project with SODIUM_USE_PKG_CONFIG.

adamreichold:
> Try to build the project with SODIUM_USE_PKG_CONFIG.

I did build it against libsodium 1.0.18 installed on my system without issues which is what I expected as I did not change the bindings after all. I think we will only get into trouble if `stable` will ever make API changes that are incompatible with the latest tagged release which should not happen as I understand the libsodium documentation.

We just need to make sure that we always keep the update priorities and do one of:
* Update only our submodule to the latest `stable` branch.
* Update our binaries from the latest `stable` snapshot and update our submodule.
* Update our bindings to the latest point release and our update the prebuilt binaries and update our  submodule.

This way we'll always end up with submodule >= prebuilt binaries >= expected API corresponding to tagged release.

Also note that there is precedence of doing this way of which I only became aware yesterday, i.e. [jedisct1/libsodium-sys-stable](https://github.com/jedisct1/libsodium-sys-stable).

kpp:
The question is: whether libsodium sets its version to stable in pkgconfig?

adamreichold:
> The question is: whether libsodium sets its version to stable in pkgconfig?

I don't think this is relevant here: The two versions involved are 
- the `VERSION` constant we manually define to be `1.0.18` in `build.rs`
- the version reported by the user's `pkg-config` manifest which will be a tagged release like `1.0.18`

Or put differently: Just because our internal snapshot is from the stable branch does not in any affect what the system `pkg-config` installation reports or the value we expect via the `VERSION` constant.

The `VERSION` constant will essentially mean the API version we expect and our internal snapshot will be never than that tagged release but will have the same API.

kpp:
> I don't think this is relevant here

It is. If a user wants to use `SODIUM_USE_PKG_CONFIG` and installs the latest libsodium via system package manager, they want our build system to use it.

Right now I got:

```
$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/libsodium.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: libsodium
Version: 1.0.17
Description: A modern and easy-to-use crypto library

Libs: -L${libdir} -lsodium
Cflags: -I${includedir}
```

Which is fine. I would like it to be 1.0.18 but I don't have updates on my machine. So I got warning:

```
warning: Using libsodium version 1.0.17.  This crate is tested against 1.0.18 and may not be fully compatible with 1.0.17.
```

If a user will get the latest updates for libsodium, will libsodium.pc contain `Version: stable`?

adamreichold:
> If a user will gets the latest updates for libsodium, will libsodium.pc contain Version: stable?

It depends on the distribution but I suspect that it will continue to contain 1.0.18 as the stable branch should be fully incompatible. If they build from source using e.g. https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable.tar.gz then libsodium.pc will indeed stay exactly at 1.0.18.

adamreichold:
> then libsodium.pc will indeed stay exactly at 1.0.18.

The same is also true when building a Git checkout of the stable branch, i.e. the pkg-config version is the same as the last tagged release.

adamreichold:
> If a user will gets the latest updates for libsodium, will libsodium.pc contain Version: stable?

What I still do not understand is how this relates to the commit we point our submodule at?

kpp:
Because right now we both rely on version = "1.0.18" in our build.rs and in sodium.pc. I am not quite sure whether there could be a new libsodium release  "1.0.19" under "stable" tag.

adamreichold:
> Because right now we both rely on version = "1.0.18" in our build.rs and in sodium.pc. I am not quite sure whether there could be a new libsodium release "1.0.19" under "stable" tag.

I doubt it as stable is described as "important fixes since the last tagged release". But in any case, it would not be a problem if our submodule follows the stable branch and the stable branch bumps it version (as long as it stays compatible): We do _not_ check the pkg-config version of the libsodium built from source, we _only_ check the pkg-config version if link against the system library. So even if our snapshot says version 2.3.4 in its libsodium.pc, as long as our `const VERSION: &str = "1.0.18";` is unchanged, the checks against the system library are unaffected.

jedisct1:
`stable` will become based on `1.0.19` once the `1.0.19` point release has been tagged.

But the `1.0.18-stable` distfiles will still be available, and security fixes will be backported there if needed.

The version number never changes within a release life cycle, including in `pkg-config`. The API doesn't change, and functions are expected to still behave as documented.

adamreichold:
> `stable` will become based on `1.0.19` once the `1.0.19` point release has been tagged.

And at this point, we will update the submodule, the binaries we ship and the expected pkg-config version instead of just updating the submodule.

kpp:
@jedisct1 

> stable will become based on 1.0.19 once the 1.0.19 point release has been tagged.

Does it mean that our warnings

`warning: Using libsodium version 1.0.18.  This crate is tested against stable and may not be fully compatible with 1.0.18.`

or 

`warning: Using libsodium version 1.0.19.  This crate is tested against stable and may not be fully compatible with 1.0.19.`

will be useless?

adamreichold:
> Does it mean that our warnings

We do not produce such warnings even with this MR. To do so we would need to change `const VERSION: &str = "stable";` which would not be useful IMHO, we just need to bump `VERSION` as we update our stable snapshot past a point where

> stable will become based on 1.0.19 once the 1.0.19 point release has been tagged.

but that will a manual procedure involving a commit and code review as every update of the submodule is.

kpp:
I see.

kpp:
> We should probably decide on

What about it?

adamreichold:
> > We should probably decide on
> 
> What about it?

> How to handle the binaries we ship for MinGW and MSVC as they are from the tagged release.

I think we should the libraries from the snapshot builds, e.g. https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip. This means they could be lagged behin the submodule and could be newer than the tagged point release we check for. Agreed?

> How to handle the version check we building against an external libsodium version.
before merging this.

We should keep the version check as it is now, i.e. we just continue to check against 1.0.18. When a new tagged release of libsodium is made, we bump the submodule, exchange the binaries and modify the version check. Agreed?

kpp:
Yes.

Update this PR to the latest master.

adamreichold:
> Yes.

Update the pull request cover letter to summarize the decisions.

> Also update this PR to the latest master.

Done.

kpp:
But you did not update our binary blobs to https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip. Why?

adamreichold:
> But you did not update our binary blobs to https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip. Why?

Only as I did not want to enlarge the scope of this pull request. But I'll add a commit...

adamreichold:
> > But you did not update our binary blobs to https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip. Why?
> 
> Only as I did not want to enlarge the scope of this pull request. But I'll add a commit...

Done. Could you please verify that the binaries I checked in match what you can download from https://download.libsodium.org/libsodium/releases/, archives libsodium-1.0.18-stable-mingw.tar.gz and libsodium-1.0.18-stable-msvc.zip? Thanks!

adamreichold:
(Note that I also bumped the libsodium snapshot by two commits to https://github.com/jedisct1/libsodium/commit/b69159b5f89b95930fb162a75bb3854f62eb536a)

kpp:
> Could you please verify 

Confirmed

kpp:
bors: r+

PR #401, opened 2019-12-04
Use cc crate to determine CFLAGS for building libsodium

adamreichold:
Determine the `CFLAGS` passed to libsodium's configure script via the `cc` crate automatically picks up environment variables like `CFLAGS` and `TARGET_CFLAGS`, handles the differences between
release and debug builds and finally provides `CFLAGS` that interoperate well when linking the result into a Rust binary by requesting per-function sections for example.

It is also more consistent with how we already determine the compiler command itself via for example consider `TARGET_CC` whereas our own `CFLAGS` logic did not consider `TARGET_CFLAGS`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
     script:
       - cargo install cargo-travis --debug || echo "cargo-travis has been already installed"
       - mkdir target # fix for cargo-coveralls
-      - cargo coveralls
+      - cargo coveralls --exclude-pattern "$PWD/target/"
 
 script:
   - cargo build --verbose
--- a/libsodium-sys/build.rs
+++ b/libsodium-sys/build.rs
@@ -152,10 +152,9 @@ fn make_libsodium(target: &str, source_dir: &Path, install_dir: &Path) -> PathBu
     use std::{fs, process::Command, str};
 
     // Decide on CC, CFLAGS and the --host configure argument
-    let build = cc::Build::new();
-    let mut compiler = build.get_compiler().path().to_str().unwrap().to_string();
-    let mut cflags = env::var("CFLAGS").unwrap_or(String::default());
-    cflags += " -O2";
+    let build_compiler = cc::Build::new().get_compiler();
+    let mut compiler = build_compiler.path().to_str().unwrap().to_string();
+    let mut cflags = build_compiler.cflags_env().into_string().unwrap();
     let host_arg;
     let cross_compiling;
     let help;


.travis.yml
```diff
@@ -41,7 +41,7 @@ matrix:
     script:
       - cargo install cargo-travis --debug || echo "cargo-travis has been already installed"
       - mkdir target # fix for cargo-coveralls
-      - cargo coveralls
+      - cargo coveralls --exclude-pattern "$PWD/target/"
```
kpp: What's that for?

kpp:
On my machine both debug and release builds take ~35s. Why?

libsodium-sys/build.rs
```diff
@@ -152,10 +152,9 @@ fn make_libsodium(target: &str, source_dir: &Path, install_dir: &Path) -> PathBu
     use std::{fs, process::Command, str};
 
     // Decide on CC, CFLAGS and the --host configure argument
-    let build = cc::Build::new();
-    let mut compiler = build.get_compiler().path().to_str().unwrap().to_string();
-    let mut cflags = env::var("CFLAGS").unwrap_or(String::default());
-    cflags += " -O2";
+    let build_compiler = cc::Build::new().get_compiler();
+    let mut compiler = build_compiler.path().to_str().unwrap().to_string();
+    let mut cflags = build_compiler.cflags_env().into_string().unwrap();
```
kpp: Why .unwrap() instead of .unwrap_or(..)?

.travis.yml
adamreichold: The coverage builds enable instrumentation via compiler flags which are now forwarded to the build of libsodium. Hence the source code of libsodium is included in the coverage metrics which made them drop by more than 50%. This excludes the libsodium code under `target/debug/libsodium-sys-*/out` from the coverage computation as we are testing the bindings assuming the libsodium implementation to be correct.

libsodium-sys/build.rs
adamreichold: `.unwrap()` asserts the validity of the conversion from `OsString` to `String` similar to `path().to_str().unwrap()` above. Generally, the build script seems to assume that paths and OS strings are valid UTF-8 and uses `unwrap` where this needs to be asserted. So I tried to keep things consistent and also to avoid silently dropping user-supplied flags if there are not valid UTF-8.

adamreichold:
> On my machine both debug and release builds take ~35s. Why?

I'd say this is hard to answer without access to the machine in question. For example, on my machine, debug builds are even consistently _slower_ than release builds.

If I'd be force to speculate, I'd note that C is relatively fast to compile so this build might actually be I/O bound more often than it is CPU bound. Release builds will spend more CPU time optimizing the generated code, but debug builds will spend more time writing out debug information like symbol names and locations. So for example, my machine might do better on CPU than IO and hence even favor release over debug builds. 

In my personal opinion, a consistent build is often more important the sheer build time, e.g. being able to single step from my Rust binary down into the libsodium implementation of some high-level could be very helpful in diagnosing thorny issues. Passing instrumentation flags for test coverage or performance analysis down to the code can be beneficial as well. In the long run, using `cc` to pass consistent compiler flags could make things like [cross language LTO](https://doc.rust-lang.org/rustc/linker-plugin-lto.html) happen automatically when a compatible C toolchain is detected.

kpp:
bors: r+

PR #403, opened 2019-12-05
Let Cargo capture subcommand output

adamreichold:
Instead of explicitly capturing the output of the subcommands we invoke in our build script, we simply let them inherit our standard input and output so that Cargo's usual logic of capturing
the output of build scripts and only displaying it if the build script failed applies.

This means that one can use `cargo build -vvv` to get the full build even if it does not fail.


kpp:
Does a release build adds O2 cflags?

adamreichold:
> Does a release build add O2 cflags?

I suspect this question is related to #401? And yes, release builds are done with optimization even though `cargo` resp. `rustc` and hence `cc` currently default to `-O3` instead of `-O2`

kpp:
bors: r+

PR #404, opened 2019-12-05
Disable bindgen layout consistency tests as they are not portable

adamreichold:
If we want to test all crates on differently wide platforms in our CI, we need to avoid bindgen's layout consistency tests as they not portable between architectures even if the bindings are.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
     script:
       - cargo install cargo-travis --debug || echo "cargo-travis has been already installed"
       - mkdir target # fix for cargo-coveralls
-      - cargo coveralls --exclude-pattern "$PWD/target/"
+      - cargo coveralls --exclude-pattern "$PWD/target/,$PWD/libsodium-sys/src/sodium_bindings.rs"
 
 script:
   - cargo build --verbose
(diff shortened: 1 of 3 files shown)


adamreichold:
@kpp I am not sure how to handle the coverage fallout here. The change is real but the coverage implied by the layout consistency tests is rather shallow IMHO and I think this is required to resolve #346.

kpp:
Woah. I will check all your PRs tomorrow. I am on a flight.

kpp:
> I am not sure how to handle the coverage fallout here.

Was it a bug in coveralls.io?

adamreichold:
> > I am not sure how to handle the coverage fallout here.
> 
> Was it a bug in coveralls.io?

No, the layout tests do "cover" the bindings generated by `bindgen`. So either we accept it and somehow override the Coveralls job or we add another exclude for the `sodium_bindings.rs` as they are automatically generated anyway.

kpp:
But the first coverage was: `Overall coverage decreased (-15.6%) to 79.63%` however the current one is `Overall coverage remained the same at 95.184%`

adamreichold:
> But the first coverage was: `Overall coverage decreased (-15.6%) to 79.63%` however the current one is `Overall coverage remained the same at 95.184%`

I amended the exclude pattern as there really is little point in computing test coverage for automatically generated code.

kpp:
bors: r+

PR #405, opened 2019-12-06
Make upcoming Clippy 1.40 happy

adamreichold:
The Clippy version shipped with the upcoming 1.40 release will have a few new complaints about our build scripts which this commit fixes.


kpp:
bors: r+

PR #406, opened 2019-12-07
RFC: Try to simplify build options for libsodium

### LINKED ISSUE #359: unify environment variables with those used in pkg-config
Instead of `SODIUM_SHARED`, why not unify the environment variables with those as-used by  `pkg-config`? Check for `SODIUM_DYNAMIC` or `SODIUM_STATIC`.

Also, why not default to using `pkg-config` and fall back to explicit handling if it's not found there (or if environment disabling it is provided)? I'm including `libsodium-sys` as part of the tests for a library I'm working on, and rebuilding libsodium from scratch every time nightly is updated is frustrating. I can specify the environment variable every time, but that shouldn't be required.

I'm happy to do the work here, just want to know if such a PR would be accepted:

* default to use pkg-config
  * probe manually if `SODIUM_LIB_DIR` is provided
  * skip if `SODIUM_NO_PKG_CONFIG` is provided
* default to static linking
  * if `SODIUM_DYNAMIC` is provided, use dynamic linking
  * change documentation to refer to `SODIUM_DYNAMIC`
  * deprecate `SODIUM_SHARED` but still allow it for now

### LINKED ISSUE #377: libsodium compilation. We do it wrong.

adamreichold:
* Make sure we use the target when deciding on what to probe, link or build.
* Make it easy to persistently enable pkg-config and vcpkg via feature flag.
* Align environment variables with pkg-config and vcpkg bindings.

Closes #377 
Closes #359 

This is my first try that does _not_ take any precautions w.r.t. backwards compatibility but sticks to the idea that the bundled build is the default while making it easy to persistently enable pkg-config or vcpkg usage. I also separated linking prebuilt libraries and building from source as this simplifies building from source a bit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,9 @@ serde_json = "^1.0.17"
 rmp-serde = "^0.13.7"
 
 [features]
+default = ["serde", "std", "bundled"]
 benchmarks = []
 std = []
-default = ["serde", "std"]
-use-pkg-config = ["libsodium-sys/use-pkg-config"]
+bundled = ["libsodium-sys/cc"]
+use-pkg-config = ["libsodium-sys/pkg-config"]
+use-vcpkg = ["libsodium-sys/vcpkg"]
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,7 +21,6 @@ environment:
   - TARGET: x86_64-pc-windows-msvc
     RUST_CHANNEL: stable
     VCPKG_TARGET: x64-windows
-    SODIUM_USE_PKG_CONFIG: 1
     VCPKGRS_DYNAMIC: 1
     RUST_BACKTRACE: 1
 
@@ -33,7 +32,7 @@ install:
       cmd.exe /c .\rustup-init.exe -y --profile minimal --default-host "$env:TARGET" --default-toolchain "$env:RUST_CHANNEL" 2`>`&1
       $env:PATH += ";$env:USERPROFILE\.cargo\bin"
 
-      If ($env:SODIUM_USE_PKG_CONFIG -eq '1') {
+      If ($env:VCPKG_TARGET -ne '') {
         # Install libsodium via vcpkg
         vcpkg install "libsodium:$env:VCPKG_TARGET"
       }
@@ -46,5 +45,5 @@ cache:
 
 build: false
 test_script:
-  - cargo test --all --target %TARGET%
-  - cargo test --all --release --target %TARGET%
+  - cargo test --all --features use-vcpkg --target %TARGET%
+  - cargo test --all --features use-vcpkg --release --target %TARGET%
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ matrix:
 
 script:
   - cargo test --all --verbose
-  - cargo test --all --verbose --no-default-features --features std
+  - cargo test --all --verbose --no-default-features --features std --features bundled
   - cargo doc
 
 branches:
--- a/README.md
+++ b/README.md
@@ -28,13 +28,6 @@ Rust >= 1.36.0 is required because of mem::MaybeUninit.
 
 ## Basic usage
 
-### Cloning
-```
-git clone https://github.com/sodiumoxide/sodiumoxide.git
-cd sodiumoxide
-git submodule update --init --recursive
-```
-
 ### Building
 ```
 cargo build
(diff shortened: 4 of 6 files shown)


Cargo.toml
```diff
@@ -34,7 +34,9 @@ serde_json = "^1.0.17"
 rmp-serde = "^0.13.7"
 
 [features]
+default = ["serde", "std", "bundled"]
 benchmarks = []
 std = []
-default = ["serde", "std"]
-use-pkg-config = ["libsodium-sys/use-pkg-config"]
+bundled = ["libsodium-sys/cc"]
```
kpp: I want the first version of this issue to be a simple as possible and with backward compatibility. I don't see any reason why we should change our env flags to features. Could you explain why do you want to do that?

adamreichold: > I want the first version of this issue to be a simple as possible and with backward compatibility.

I can certainly keep things fully compatible. I just wanted to start with the untwisted version to the start the discussion on what we would like to have if there were no backwards compatibility concerns. After we clarified that, we can mechanically infer what can we can have with in a compatible way or how much compatibility we are going to loose in exchange for some features.

> I don't see any reason why we should change our env flags to features. 

Environment variables need to be specified every time a build is invoked as @stouset pointed out in #359 as Cargo has no way to specify them persistently. Feature flags however can be specified via a dependent crates Cargo.toml to easily and persistently set a preference for linking with the system library.

adamreichold: (For example, the only solution to #359 that is compatible that I can think is accepting both which is messy for people starting to use the crate but can be easily added here.)

kpp: sodiumoxide is used as a library which means that it can be used multiple times in a binary project which can cause `features` interference. 

> Environment variables need to be specified every time a build is invoked as

People may use `export ENV=value` to set env variable which will live until you `unset` it or end the console session

adamreichold: > sodiumoxide is used as a library which means that it can be used multiple times in a binary project which can cause `features` interference.

Yes, features need to be additive. But I tried to preserve this here, i.e. enabling a feature will enable one more way we search for libsodium with system-wide taking precedence as this is what needs to be requested explicitly.

> > Environment variables need to be specified every time a build is invoked as
> 
> People may use `export ENV=value` to set env variable which will live until you `unset` it or end the console session

For a single console session, this helps. But one still needs documentation or script to set up the initial session. This gets more cumbersome if it isn't the top-level crate the requests this. (Both pkg-config/vcpkg provide variables to suppress them on per-library basis which is useful to counteract additive features which seems to mesh well.)

As with the environment variable naming, we can probably support both if we want to (we have a pkg-config feature already), i.e. triggering pkg-config/vcpkg via feature flags and/or via an environment variable.

kpp: How about fixing `#cfg` issues and then deciding what to do with env vs features in another PR?

adamreichold: > How about fixing `#cfg` issues and then deciding what to do with env vs features in another PR?

I can certainly open another pull request try to just fix using up `#[cfg(...)]` usage but that is made difficult because frankly speaking the current `build.rs` is a bit of a mess. Hence the desire to clean up so that the different ways of linking libsodium are spelled our clearly and aligned with where the crates ecosystem as a whole seems to be going.

kpp: > I can certainly open another pull request

Yes please. This PR is too big and can cause unexpected bugs if not reviewed properly.

adamreichold: > Yes please. This PR is too big and can cause unexpected bugs if not reviewed properly.

Will do. But I admittedly cannot give you a time frame ATM.

kpp:
Hello! How is it going?

adamreichold:
> Hello! How is it going?

Nothing to report for now. But thanks for the ping. I will have this on top of my list the next time I have some spare time for OSS.

stouset:
Out of curiosity, once this is merged, is there any desire to move to a flow where sodiumoxide attempts to link to system libraries by default rather than building libsodium from scratch?

adamreichold:
The system-deps approach appears to be a more fruitful endeavor.

PR #407, opened 2019-12-08
Fix xcopy invocation

adamreichold:
When switching from robocopy to xcopy, I did not adjust the flags correctly. I try to fix this here and cherry-picked @jean-airoldie's change to verify that it actually worked (as I have no Windows machine at hand).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -44,7 +44,6 @@ cache:
   - target
 
 build: false
-
 test_script:
-  - cargo test --target %TARGET%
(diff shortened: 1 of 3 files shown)


adamreichold:
There actually was another issue: In contrast to Travis CI, AppVeyor does not automatically clone submodules which I added in addition to fixing the xcopy flags. (I also took the liberty to use rustup's minimal profile to avoid installing the Rust documentation for the CI builds.)

kpp:
bors: r+

PR #418, opened 2020-05-30
Add support for the AES256-GCM AEAD construction

klittlepage:
This commit adds support for AES256-GCM. NB: libsodium limits its support
of AES256-GCM to x86/x86-64 processors with the AES-NI instruction set.
Compile time feature flags are used to hide GCM functionality for targets
lacking the requisite support. In the interest of maintaining parity with
the libsodium API an `is_available` function mapping to
`crypto_aead_aes256gcm_is_available` is unconditionally exported but
shouldn't be required given the compile-time check.

Signed-off-by: Kelly Littlepage <kelly@onechronos.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,24 @@ Documentation will be generated in target/doc/...
 Most documentation is taken from NaCl, with minor modification where the API
 differs between the C and Rust versions.
 
+### Support for the AES AEAD Variant
+
+The AES AEAD variant `crypto_aead_aes256gcm` requires hardware support for the
+`AES` and `CLMUL` instruction set extensions to x86; you can read why that's the
+case
+[here](https://doc.libsodium.org/secret-key_cryptography/aead/aes-256-gcm#limitations). These instruction set extensions were first made
+available in Intel Westmere (early 2010) and at the time of writing x86 hardware
+support for them is near universal.
+
+Libsodium exposes an API for runtime feature detection and doesn't prevent
+you from calling `crypto_aead_aes256gcm` on a machine lacking `AES` and
+`CMUL` expressions; doing so will result in a runtime `SIGILL` (illegal
+instruction). By contrast sodiumoxide exposes an API that precludes the use of
+the `crypto_aead_aes256gcm_*` family of functions without performing the runtime
+check. It's important to note that the use of `sodiumoxide::init()` is mandatory
+when using AES; unless you call `init` calls `aead::aes256gcm::Aes256Gcm::new()`
+will always return `Err(_)` even if your runtime hardware supports AES.
+
 ## Dependencies
 
 C compiler (`cc`, `clang`, ...) must be installed in order to build libsodium from source.
--- a/src/crypto/aead/aead_macros.rs
+++ b/src/crypto/aead/aead_macros.rs
@@ -4,7 +4,8 @@ macro_rules! aead_module (($seal_name:ident,
                            $open_detached_name:ident,
                            $keybytes:expr,
                            $noncebytes:expr,
-                           $tagbytes:expr) => (
+                           $tagbytes:expr,
+                           $test_gate:expr) => (
 
 #[cfg(not(feature = "std"))] use prelude::*;
 use libc::c_ulonglong;
@@ -166,6 +167,9 @@ mod test_m {
     #[test]
     fn test_seal_open() {
         use randombytes::randombytes;
+        if !$test_gate {
+          return
+        }
         for i in 0..256usize {
             let k = gen_key();
             let n = gen_random_nonce();
@@ -180,6 +184,9 @@ mod test_m {
     #[test]
     fn test_seal_open_tamper() {
         use randombytes::randombytes;
+        if !$test_gate {
+          return
+        }
         for i in 0..32usize {
             let k = gen_key();
             let n = gen_random_nonce();
@@ -204,6 +211,9 @@ mod test_m {
     #[test]
     fn test_seal_open_detached() {
         use randombytes::randombytes;
+        if !$test_gate {
+          return
+        }
         for i in 0..256usize {
             let k = gen_key();
             let n = gen_random_nonce();
@@ -219,6 +229,9 @@ mod test_m {
     #[test]
     fn test_seal_open_detached_tamper() {
         use randombytes::randombytes;
+        if !$test_gate {
+          return
+        }
         for i in 0..32usize {
             let k = gen_key();
             let n = gen_random_nonce();
@@ -249,6 +262,9 @@ mod test_m {
     #[test]
     fn test_seal_open_detached_same() {
         use randombytes::randombytes;
+        if !$test_gate {
+          return
+        }
         for i in 0..256usize {
             let k = gen_key();
             let n = gen_random_nonce();
--- a/src/crypto/aead/aes256gcm.rs
+++ b/src/crypto/aead/aes256gcm.rs
@@ -1,0 +1,235 @@
+//! WARNING: Despite being the most popular AEAD construction due to its use in
+//! TLS, safely using AES-GCM in a different context is tricky. No more than
+//! ~ 350 GB of input data should be encrypted with a given key. This is for
+//! ~ 16 KB messages - actual figures vary according to message sizes.
+//! In addition, nonces are short and repeated nonces would totally destroy the
+//! security of this scheme. Nonces should thus come from atomic counters,
+//! which can be difficult to set up in a distributed environment.
+//! Unless you absolutely need AES-GCM, use the
+//! [default AEAD export](crate::crypto::aead)
+//! instead - it doesn't have any of these limitations. Or, if you don't need
+//! to authenticate additional data, just stick to [secretbox](crate::crypto::secretbox).
+//!
+//! AES primitives will not be made available unless your runtime CPU
+//! is x86/x86_64 with support for the AES-NI instruction set and the CLMUL
+//! instruction (Westmere and beyond).
+
+mod aes_impl {
+    use ffi::{
+        crypto_aead_aes256gcm_ABYTES, crypto_aead_aes256gcm_KEYBYTES,
+        crypto_aead_aes256gcm_NPUBBYTES, crypto_aead_aes256gcm_decrypt,
+        crypto_aead_aes256gcm_decrypt_detached, crypto_aead_aes256gcm_encrypt,
+        crypto_aead_aes256gcm_encrypt_detached,
+    };
+
+    /// `is_available` returns true if the current CPU supports aes256gcm and false otherwise.
+    ///
+    /// WARNING: You must call [init](crate::init) before calling this function; if you do not
+    /// `is_available` will always return false even if the runtime supports aes256gcm.
+    pub fn is_available() -> bool {
+        unsafe { ffi::crypto_aead_aes256gcm_is_available() == 1 }
+    }
+
+    aead_module!(
+        crypto_aead_aes256gcm_encrypt,
+        crypto_aead_aes256gcm_decrypt,
+        crypto_aead_aes256gcm_encrypt_detached,
+        crypto_aead_aes256gcm_decrypt_detached,
+        crypto_aead_aes256gcm_KEYBYTES as usize,
+        crypto_aead_aes256gcm_NPUBBYTES as usize,
+        crypto_aead_aes256gcm_ABYTES as usize,
+        crate::init().is_ok() && is_available()
+    );
+
+    #[cfg(test)]
+    mod test {
+        use super::*;
+
+        #[test]
+        fn test_vector_1() {
+            // Test vector from https://tools.ietf.org/html/rfc7714#section-16.2.2
+            let m = &[
+                0x47, 0x61, 0x6c, 0x6c, 0x69, 0x61, 0x20, 0x65, 0x73, 0x74, 0x20, 0x6f, 0x6d, 0x6e,
+                0x69, 0x73, 0x20, 0x64, 0x69, 0x76, 0x69, 0x73, 0x61, 0x20, 0x69, 0x6e, 0x20, 0x70,
+                0x61, 0x72, 0x74, 0x65, 0x73, 0x20, 0x74, 0x72, 0x65, 0x73,
+            ];
+            let ad = &[
+                0x80, 0x40, 0xf1, 0x7b, 0x80, 0x41, 0xf8, 0xd3, 0x55, 0x01, 0xa0, 0xb2,
+            ];
+            let k = Key([
+                0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d,
+                0x0e, 0x0f, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19, 0x1a, 0x1b,
+                0x1c, 0x1d, 0x1e, 0x1f,
+            ]);
+            let n = Nonce([
+                0x51, 0x75, 0x3c, 0x65, 0x80, 0xc2, 0x72, 0x6f, 0x20, 0x71, 0x84, 0x14,
+            ]);
+            let c_expected = &[
+                0x32, 0xb1, 0xde, 0x78, 0xa8, 0x22, 0xfe, 0x12, 0xef, 0x9f, 0x78, 0xfa, 0x33, 0x2e,
+                0x33, 0xaa, 0xb1, 0x80, 0x12, 0x38, 0x9a, 0x58, 0xe2, 0xf3, 0xb5, 0x0b, 0x2a, 0x02,
+                0x76, 0xff, 0xae, 0x0f, 0x1b, 0xa6, 0x37, 0x99, 0xb8, 0x7b, 0x7a, 0xa3, 0xdb, 0x36,
+                0xdf, 0xff, 0xd6, 0xb0, 0xf9, 0xbb, 0x78, 0x78, 0xd7, 0xa7, 0x6c, 0x13,
+            ];
+            let c = seal(m, Some(ad), &n, &k);
+            assert_eq!(&c[..].len(), &c_expected[..].len());
+            assert_eq!(&c[0..44], &c_expected[0..44]);
+        }
+    }
+}
+
+mod aes_api {
+    use super::aes_impl;
+    use crypto::nonce::gen_random_nonce;
+
+    /// The Aes256Gcm struct encapsulates the crypto_aead_aes256gcm_* family of
+    /// functions in a way that ensures safe usage of the API at runtime
+    /// without incurring a per function call cost.
+    #[derive(Debug, Clone, Copy)]
+    pub struct Aes256Gcm;
+
+    impl Aes256Gcm {
+        /// Returns an `Ok` of [Aes256Gcm](self::Aes256Gcm) if the runtime
+        /// supports AES and an `Err(_)` if it does not.
+        ///
+        /// You must call [init](crate::init) before calling this function. Failure
+        /// to do so will result in `Err(_)` being returned even if the runtime
+        /// hardware supports AES.
+        pub fn new() -> Result<Self, ()> {
+            if unsafe { ffi::crypto_aead_aes256gcm_is_available() } == 1 {
+                Ok(Self)
+            } else {
+                Err(())
+            }
+        }
+
+        /// `gen_initial_nonce` randomly generates an initial nonce
+        ///
+        /// WARNING: AES nonces are short enough that the probability of collision between two randomly
+        /// generated nonces is nonnegligible and repeated nonce use will totally destroy the security
+        /// of this scheme. Use [increment_le]( aes_impl::Nonce::increment_le) or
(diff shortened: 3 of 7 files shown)


klittlepage:
Hi, checking in on this. Is there any reason to not merge?

src/crypto/aead/aes256gcm.rs
```diff
@@ -0,0 +1,127 @@
...
+    )
+))]
+pub use self::aes::*;
+
+/// `is_available` returns true if the current CPU supports aes256gcm and false otherwise.
+///
+/// WARNING: You must call [init](crate::init) before calling this function; if
+/// you do not the result of calling `is_available` is undefined.
+pub fn is_available() -> bool {
+    unsafe { ffi::crypto_aead_aes256gcm_is_available() == 1 }
```
kpp, line 112: Will libsodium fallback to a software implementation if there is no pclmulqdq&aes?

klittlepage: No, there's no safe way to implement AES without hardware support, and thankfully libsodium doesn't try. I exposed the is_available check outside of the feature flag gated module to maintain API parity with libsodium. However, the compile time check alone ensures correct API usage and precludes the possibility of a runtime fault.

kpp:
```
$ rg pclmulqdq\|aes -c /proc/cpuinfo
8
$ cargo test aes
    Finished test [unoptimized + debuginfo] target(s) in 0.03s
     Running target/debug/deps/sodiumoxide-e35cd2b3f289092a

running 1 test
test crypto::aead::aes256gcm::test::test_is_available ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 237 filtered out
```

Which means I don't have access to aes256-gcm even if I have aes&pclmulqdq

klittlepage:
`RUSTFLAGS="-C target-cpu=native" cargo test aes`

```
running 7 tests
test crypto::aead::aes256gcm::aes::test::test_vector_1 ... ok
test crypto::aead::aes256gcm::test::test_is_available ... ok
test crypto::aead::aes256gcm::aes::test_m::test_seal_open_detached_tamper ... ok
test crypto::aead::aes256gcm::aes::test_m::test_seal_open_tamper ... ok
test crypto::aead::aes256gcm::aes::test_m::test_seal_open_detached ... ok
test crypto::aead::aes256gcm::aes::test_m::test_seal_open ... ok
test crypto::aead::aes256gcm::aes::test_m::test_seal_open_detached_same ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 237 filtered out
```

Rust won't use specialized instructions in test profile builds unless you either target a specific CPU or test a release profile.

Those instructions exist in Westmere and beyond (~2009) so support on x86 is near universal.

kpp:
I think it would be useful to add these env vars into CI to build your code. Also we can add rustflags to package metadata(https://docs.rs/about) so that docs.rs could render aes module in the docs.

Also I am curious whether we should add docs into README how to turn on aes.

Cargo.toml
```diff
@@ -38,3 +38,7 @@ benchmarks = []
 std = []
 default = ["serde", "std"]
 use-pkg-config = ["libsodium-sys/use-pkg-config"]
+
+[package.metadata.docs.rs]
+default-target = "x86_64-unknown-linux-gnu"
+rustc-args = ["-C target-feature=+aes", "-C target-feature=+pclmulqdq"]
```
kpp, line 44: Wouldn't -C target-cpu=native be enough?

kpp:
LGTM. Does anybody want to review this PR?

Cargo.toml
klittlepage: To your earlier point doing it this way ensures that the docs will always be “expanded” regardless of what platform they’re built on. I could arbitrarily pick a target with AES/CMUL support but doing it this way documents intent. There’s an [open discussion](https://github.com/rust-lang/rust/issues/1998) about how rude docs should handle conditional compilation (feature detection or otherwise) but this seems like the lesser of evils for now.

README.md
```diff
@@ -55,6 +55,33 @@ Documentation will be generated in target/doc/...
 Most documentation is taken from NaCl, with minor modification where the API
 differs between the C and Rust versions.
 
+#### Building and Testing with Support for AES
+
+The AES AEAD variant `crypto_aead_aes256gcm` requires hardware support for the
+`AES` and `CLMUL` instruction set extensions to x86; you can read why that's the
+case
+[here](https://libsodium.gitbook.io/doc/secret-key_cryptography/aead/aes-256-gcm#limitations). These instruction set extensions were first made
```
adamreichold, line 63: Maybe prefer the link that includes the navigation for context, e.g. https://doc.libsodium.org/secret-key_cryptography/aead/aes-256-gcm#limitations

README.md
```diff
@@ -55,6 +55,33 @@ Documentation will be generated in target/doc/...
...
+available in Intel Westmere (early 2010) and at the time of writing x86 hardware
+support for them is near universal. However, rustc is a cross compiler and will
+not enable them unless it's targeting a CPU that supports them; the baseline
+x86 target does not.
+
+If you need to use `crypto_aead_aes256gcm` you must build/test sodiumoxide with
+`RUSTFLAGS` set to include one of the following option sets:
+
+* `-C target-cpu=native` if you're building on a box with `AES` and `CLMUL`
+* `-C target-cpu=[target]` where `[target]` is `rustup` target with `AES` and
```
adamreichold, line 73: I don't think this adds up, a `rustup` target, i.e. something listed by `rustup target list`, is a host-tuple that is passed via `--target` to rustc whereas the possible options for the target CPU are given by `rustc --print target-cpus`. (It is also not a target in itself but just an alias for a collection of target features AFAIK.)

src/crypto/aead/aes256gcm.rs
```diff
@@ -0,0 +1,127 @@
...
+//! [default AEAD export](crate::crypto::aead)
+//! instead - it doesn't have any of these limitations. Or, if you don't need
+//! to authenticate additional data, just stick to [secretbox](crate::crypto::secretbox).
+//!
+//! AES primitives will not be made available unless your compile target
+//! is x86/x86_64 with support for the AES-NI instruction set and the CLMUL
+//! instruction (Westmere and beyond).
+
+#[cfg(any(
+    doc,
```
adamreichold, line 18: Do we need the `any(doc, ...)` arm here now that the flags for `docs.rs` are adjusted? (Running `cargo doc` should produce whatever your compiler flags will produce for the, shouldn't it?)

README.md
```diff
@@ -55,6 +55,33 @@ Documentation will be generated in target/doc/...
...
+`AES` and `CLMUL` instruction set extensions to x86; you can read why that's the
+case
+[here](https://libsodium.gitbook.io/doc/secret-key_cryptography/aead/aes-256-gcm#limitations). These instruction set extensions were first made
+available in Intel Westmere (early 2010) and at the time of writing x86 hardware
+support for them is near universal. However, rustc is a cross compiler and will
+not enable them unless it's targeting a CPU that supports them; the baseline
+x86 target does not.
+
+If you need to use `crypto_aead_aes256gcm` you must build/test sodiumoxide with
+`RUSTFLAGS` set to include one of the following option sets:
```
adamreichold, line 70: I would suggest avoid advocating use of the `RUSTFLAGS` enviroment variable here as in a production setting this should often be done via [Cargo configuration](https://doc.rust-lang.org/cargo/reference/config.html) instead. Maybe talk about `rustc` flags in a generic manner?

adamreichold:
> Compile time feature flags are used to hide GCM functionality for targets
lacking the requisite support.

@klittlepage One thing I do not fully understand is the interaction of rustc target features with how libsodium is compiled/does implementation selection here, i.e. doesn't libsodium do CPU feature detection at runtime, c.f. https://github.com/jedisct1/libsodium/blob/master/src/libsodium/sodium/runtime.c, and hence could have a hardware-backed AES implementation available even if Rust code was not compiled to take advantage of these instructions? Meaning that I am not really convinced that hiding that API based on the compile-time target features is helpful as they have nothing to do with whether these instructions will be available at runtime and these bindings should probably be as flexible as the underlying implementation of libsodium is.

README.md
```diff
@@ -55,6 +55,33 @@ Documentation will be generated in target/doc/...
...
+* `-C target-cpu=native` if you're building on a box with `AES` and `CLMUL`
+* `-C target-cpu=[target]` where `[target]` is `rustup` target with `AES` and
+  `CLMUL` support
+* `-C target-feature=+aes -C target-feature=+pclmulqdq`
+
+Note that in the latter two cases you must make sure that your executable runs
+on suitable hardware; attempting to run it on a processor that
+*does not support* `AES` and `CLMUL` will result in a runtime `SIGILL`
+(illegal instruction). You can test for hardware support at runtime using
+`aead::aes256gcm::is_available` (make sure that the rustoxide `init` method is
```
adamreichold, line 81: Is "rustoxide" maybe meant to be "sodiumoxide"?

README.md
klittlepage: Yup, I didn't document that correctly. The new PR clarifies.

src/crypto/aead/aes256gcm.rs
klittlepage: Unfortunately we do still need it until https://github.com/rust-lang/cargo/issues/7677 is resolved.

The docs won't build without it.

README.md
klittlepage: Oops...

klittlepage:
> > Compile time feature flags are used to hide GCM functionality for targets
> > lacking the requisite support.
> 
> @klittlepage One thing I do not fully understand is the interaction of rustc target features with how libsodium is compiled/does implementation selection here, i.e. doesn't libsodium do CPU feature detection at runtime, c.f. https://github.com/jedisct1/libsodium/blob/master/src/libsodium/sodium/runtime.c, and hence could have a hardware-backed AES implementation available even if Rust code was not compiled to take advantage of these instructions?

Yes, that's true. How the rust side of things is compiled has no impact on the compilation of libsodium and the libsodium API exports AES unconditionally.

> Meaning that I am not really convinced that hiding that API based on the compile-time target features is helpful as they have nothing to do with whether these instructions will be available at runtime and these bindings should probably be as flexible as the underlying implementation of libsodium is.

Yes, and if the goal is to mirror libsodium that makes sense. However, there are plenty of examples where sodiumoxide improves on the safety of the libsodium API e.g. the use of strongly typed keys, nonces, and tags in AEAD. These approaches aren't as "flexible" as the underlying implementation of libsodium - and that's a good thing.

Putting AES functionality behind a compile time feature check limits the possibility of a foot gun (using AES on an architecture that doesn't support it) so I'm in favor of making the behavior opt-in. I added a feature flag "aes" for users who have read the caveats and want to force an unconditional build.

kpcyrd:
I don't think testing cpu features at compile time is a good idea, instead this should be done at runtime. Binary distribution is very common and the cpu of the build system is almost guaranteed to be different from the system that's actually going to run it. A binary compiled on a cpu with support for aes instructions should be able to run on cpus without this instruction and vice versa. I've accidentally shipped broken updates to users in the past because the build system was feature-tested and the final binary assumed the instructions would be present without further runtime checks, causing SIGILL for some users.

Not testing this at compile-time is also important for reproducible builds, which relies on the build to be deterministic to allow independent verification of the binary on a system that isn't 100% identical.

I'm surprised libsodium doesn't have a fallback, does anybody know why? @jedisct1

klittlepage:
> I've accidentally shipped broken updates to users in the past because the build system was feature-tested and the final binary assumed the instructions would be present without further runtime checks, causing SIGILL for some users.

That assumes that users know how to use the API correctly in the first place first place. Libsodium doesn't gate the internals of AES with a runtime feature check - it relies on the caller calling `init` and `crypto_aead_aes256gcm_is_available`. Moving the check into the relevant functions could make for significant overhead and doesn't really get you much aside from getting an unrecoverable error instead of a signal.

Here are the compile and run scenarios for conditional compilation:

1. runtime: has AES, target: emit AES => ✅
2. runtime: has AES, target: emit AES => no aead_aes256gcm module exported => user forced to read docs and resolve => ✅
3. runtime: does not have AES, target: emit AES => ✅
4. runtime: does not have AES, target: emit AES => problem that the user explicitly opted into when cross compiling incorrectly

Without conditional compilation 3) and 4) are the same and much more likely to happen on accident. At worst, 2) forces the user to read the docs. If that's viewed as a bad thing, that really doesn't bode well for correct use of `crypto_aead_aes256gcm_is_available`.

> Not testing this at compile-time is also important for reproducible builds, which relies on the build to be deterministic to allow independent verification of the binary on a system that isn't 100% identical.

That's why - as @adamreichold noted - using rust's configuration system is the way to go. Reproducibility is always modulo the build environment; conditional compilation has no impact on reproducibility.

> I'm surprised libsodium doesn't have a fallback, does anybody know why? @jedisct1

AES in GCM is notoriously prone to timing side-channels in the computation of the Galois hash unless you have access to hardware primitives for constant time carry-less multiplication. There are software mitigants that are slower and more complex; the latter point is the real concern.

kpcyrd:
> The assumption that a runtime check resolves that issue hinges on correct API usage. Libsodium doesn't gate the internals of AES with a runtime feature check - it relies on the caller calling init and crypto_aead_aes256gcm_is_available. The user has to know to do that. Moving the check into the relevant functions could make for significant overhead and doesn't really get you much aside from an unrecoverable error instead of a signal.

Compile-time checks don't solve this. I'd argue there's (unfortunately) no way to expose `crypto_aead_aes256gcm` in a safe and performant interface due to the lack of a fallback and the only reasonable option is delegating the responsibility to check `crypto_aead_aes256gcm_is_available` to the developer consuming the sodiumoxide api.

Your argument dismisses the fact that binary distributions like Debian and Arch Linux exist, "the user" might not be the person writing the code, and the person compiling it might be yet another person. In the opensource world the binaries are deployed on a much broader set of cpus than in controlled, proprietary deployments, there is no concept of a "target" more specific than amd64/x86_64.

It's also going to cause reproducible build issues for real packages that depend on sodiumoxide today and are currently considered reproducible according to [reproducible-builds.org standards](https://github.com/bmwiedemann/theunreproduciblepackage/tree/master/compile-time-check/cpu-detection), the build machine cpu is not considered a build input.

My best guess is having `sodiumoxide::init` return an `Err(_)` if the aes feature is enabled and `crypto_aead_aes256gcm_is_available` returns false at runtime. Calling `sodiumoxide::init` isn't mandatory right now though.

adamreichold:
> Yes, and if the goal is to mirror libsodium that makes sense. However, there are plenty of examples where sodiumoxide improves on the safety of the libsodium API e.g. the use of strongly typed keys, nonces, and tags in AEAD. These approaches aren't as "flexible" as the underlying implementation of libsodium - and that's a good thing.

I am very much on board with these bindings improving the safety beyond what libsodium itself offers, I just do not really see the target features helping here as:
* If we enable the target features and run on a CPU without them, we get SIGILL.
* If we rely purely on runtime checks and fail to perform them before calling the AES functions, we get SIGILL.

Considering that the person building the software is often not the same person that is running the software, the outcome of an incorrect assumption embedded in the build config and an incorrect/missing runtime check is the same: the program will crash.

 > Moving the check into the relevant functions could make for significant overhead and doesn't really get you much aside from an unrecoverable error instead of a signal.

> Compile-time checks don't solve this. I'd argue there's (unfortunately) no way to expose crypto_aead_aes256gcm in a safe and performant interface due to the lack of a fallback and the only reasonable option is delegating the responsibility to check crypto_aead_aes256gcm_is_available to the developer consuming the sodiumoxide api.

Did we actually measure the overhead of e.g. using `std::sync::Once` to check `crypto_aead_aes256gcm_is_available` the first time before one of the `crypto_aead_aes256gcm_*` functions is called? (I think this should be safe even if `init` was not called as no CPU feature will have been detected and `crypto_aead_aes256gcm_is_available` should also return `false`.)

adamreichold:
> Did we actually measure the overhead of e.g. using std::sync::Once to check crypto_aead_aes256gcm_is_available the first time before one of the crypto_aead_aes256gcm_* functions is called? (I think this should be safe even if init was not called as no CPU feature will have been detected and crypto_aead_aes256gcm_is_available should always return false.)

@klittlepage I think we can have our cake and it eat it too: Instead of exposing free functions as an API, we could add a unit struct like `pub struct Aes256Gcm;` with a `pub fn new() -> Self` method which calls `crypto_aead_aes256gcm_is_available` (either panicking or making `new` return a `Result` instead). The actual API functions then become methods of this struct which does not impose any overhead with the zero-sized type only providing a type-level proof that the runtime check was performed.

Of course this might imply avoid the `aead_module!` macro and rests on  `crypto_aead_aes256gcm_is_available` returning `false` if `init` was never called, but it should provide a way to safely expose these functions without runtime overhead.

src/crypto/aead/aes256gcm.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+mod aes_api {
+    use super::aes_impl;
+    use crypto::nonce::gen_random_nonce;
+
+    /// The Aes256Gcm struct encapsulates the crypto_aead_aes256gcm_* family of
+    /// functions in a way that ensures safe usage of the API at runtime
+    /// without incurring a per function call cost.
+    pub struct Aes256Gcm;
```
adamreichold, line 87: I think we should derive `Debug`, `Clone` and `Copy` here to make it as easy as possible to embed and pass around this token.

.travis.yml
```diff
@@ -49,8 +49,8 @@ matrix:
       - cargo check --all --verbose
 
 script:
-  - cargo test --all --verbose
-  - cargo test --all --verbose --no-default-features --features std
+  - RUSTFLAGS="-C target-cpu=native" cargo test --all --verbose
```
adamreichold, line 52: Do we still need this? (It does a bit a variability to the build if the CPU of the CI runner changes, so I think it would be nice to avoid it if we can.)

adamreichold:
Minor nit on the trait impls of the unit struct, otherwise this looks good to me now!

klittlepage:
> @klittlepage I think we can have our cake and it eat it too: Instead of exposing free functions as an API, we could add a unit struct like `pub struct Aes256Gcm;` with a `pub fn new() -> Self` method which calls `crypto_aead_aes256gcm_is_available` (either panicking or making `new` return a `Result` instead).

If folks are open to AES having a slightly different API, this is by far the best way to do it. My earlier commit reflected an assumption that the API needed to remain consistent. I've updated the implementation and docs.

@kpcyrd to your earlier point that "Compile-time checks don't solve this" - just to clarify - I never said that they did. The argument for using them was simply that it requires someone (either the person building the package) or consuming the API to explicitly opt in to the behavior having read the docs. Your point that the value of that is lessened when different people build/package/consume the library is completely true, so the AES specific API really does seem like the best solution.

I ran a Criterion test across all of the AEAD primitives (violin plot attached). AES enjoys a slight performance lead at every payload size and a more significant reduction in variance; there's not a compelling performance reason to favor it over the stream ciphers.

aead_variants_violin

.travis.yml
kpp: How would you test aes?

src/crypto/aead/aead_macros.rs
```diff
@@ -4,7 +4,8 @@ macro_rules! aead_module (($seal_name:ident,
                            $open_detached_name:ident,
                            $keybytes:expr,
                            $noncebytes:expr,
-                           $tagbytes:expr) => (
+                           $tagbytes:expr,
+                           $test_gate:expr) => (
```
kpp, line 8: I'd rather have `target_feature = "aes"` for aes tests

.travis.yml
klittlepage: Under the new construction the check is at runtime so the AES tests will run on any hardware that supports it regardless of how the crate is compiled.

src/crypto/aead/aead_macros.rs
klittlepage: I'm not sure how that would work unless you wanted to expand the test matrix to do a separate AES and non AES build, in which case we're back to the earlier discussion about packaging and reproducibility.

kpp: Instead of fixing tests for aes only (`crate::init().is_ok() && is_available()`) you added $test_gate variable for all modules. Am I missing something?

klittlepage: The goal was to keep the unit tests consistent across all of the internal implementations of aead. The only way to do that is introduce a gate that checks preconditions (runtime support in this case). There are a few ways to factor it, but this seemed like the most minor surgery.

adamreichold: Personally, I really like this way of keeping the amount of boiler plate as small as possible without introducing any complications into the build.

The only other reasonable option I see would be to require the unit tests to be run on a machine with support for hardware AES which does sound entirely reasonable considering that they wont run with 16MB of RAM or take unpratically long on a 33 MHz CPU as well.

klittlepage: @kpp are we good on this point?

kpp: yes.

klittlepage:
@kpcyrd is anything else needed on my end?

kpp:
bors: r+

kpp:
@klittlepage rebase plz

klittlepage:
@kpp done - thank you!

kpp:
bors: r+

Issue #363, opened 2019-08-18
Support for bare-metal compilation


ryankurte:
hi hi

at the moment compilation depends on libc (and std as in #197) which is not defined for all architectures.

i've had a go at [moving from libc to cty](https://github.com/sodiumoxide/sodiumoxide/compare/master...ryankurte:fix/arm-none-eabi?expand=1) which solves part of the problem (though is v roughly implemented), though something appears to be something wrong with PIC/PIE outputting relocatable code (maybe related to https://github.com/jedisct1/libsodium/issues/677, see also: https://github.com/rust-embedded/cortex-m-rt/issues/22).

this results in: 
```
  = note: /usr/lib/gcc/arm-none-eabi/7.3.1/../../../arm-none-eabi/bin/ld: error: no memory region specified for loadable section `.got'
          collect2: error: ld returned 1 exit status
```
at the final linking stage. i'm setting `SODIUM_DISABLE_PIE=1` and hardcoding `--no-pie` and `-static` as `CFLAGS` in [libsodium-sys/src/build.rs](https://github.com/sodiumoxide/sodiumoxide/compare/master...ryankurte:fix/arm-none-eabi?expand=1#diff-03b86c4d158daab101ab51fd64460e9bR232) but i'm basically out of ideas.

as an aside, have you considered moving to building with cc instead of via autotools to simplify all this? i've done this in cmake before and it wasn't too complex (see also: https://github.com/sodiumoxide/sodiumoxide/issues/377, https://github.com/sodiumoxide/sodiumoxide/pull/379, https://github.com/sodiumoxide/sodiumoxide/pull/406)

ryankurte:
Possibly related to https://github.com/fortanix/rust-mbedtls/pull/63, no solution as yet </3

kpp:
> have you considered moving to building with cc instead of via autotools to simplify all this?

I suggest you ask jedisct1 whether it's possible

jedisct1:
[As documented](https://doc.libsodium.org/installation#cross-compiling), all you need is `--specs=nosys.specs` in the linker flags.

ryankurte:
Unfortunately this does not appear to be the case, I have fixed a bunch of the `sodiumoxide` specific issues in #422 however no combination of `LDFLAGS=--specs=nosys.specs` or `--disable-pie` appears to have any effect on the behaviour of `./configure`.

Running against #442 with:
```
[...]

PR #422, opened 2020-06-30
Fix builds for `no_std` platforms

ryankurte:
Relates to #363. This does not yet work.

#### Adds `libc` and `cty` features to `sodiumoxide` and `libsodium-sys`, removes use of `libc::strlen`

libc is not available on all `no_std` platforms.

#### Adds `SODIUM_HOST` build argument to `libsodium_sys/build.rs` 

This allows overriding of `./configure --host` as cargo `HOST` tuples do not match those expected by the `configure` script, leading to errors such as:

```
checking host system type... 
--- stderr
Invalid configuration `thumbv7em-none-eabihf': machine `thumbv7em-none' not recognized
configure: error: /bin/bash build-aux/config.sub thumbv7em-none-eabihf failed

```

#### Adds explicit `LDFLAGS` passing to configure command in `libsodium_sys/build.rs`

Afaik this should not be required, however, neither way works.


kpp:
@ryankurte ping. We fixed our CI (moved to github actions). Would you please update your PR?

ryankurte:
so this is _working_ on the `no_std` platform i'm using, but, still requires:

- `SODIUM_DISABLE_PIE=1 LDFLAGS='--specs=nosys.specs' SODIUM_HOST=arm-none-eabihf cargo build` to convince sodium of the architecture
- `"-C", "link-arg=-lc",` to include libc for sodium to link against
- Mock [newlib stubs](https://gist.github.com/ryankurte/df961a7c36b6c2c4fa13fc052e60bf0c) to keep the linker happy with libc inclusion
- `"-C", "link-arg=-L/usr/lib/arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/",` to convince the linker to use the correct libc symbols (though apparently this shouldn't be required)

klittlepage:
@ryankurte do you have documented local build and test procedures so that we could try to get it working in the CI under a `no_std` target?

kpp:
Well at least your PR should not fail other targets.

kpp:
Hello. Will you continue working on this PR?

ryankurte:
hey i'm planning to but, i have do a working build for my purposes and it already cost me a couple of weeks of working out linking etc. so, it's not super high on my list of priorities at the moment (if anyone else wants to pick it up please feel free!)

you can create a no-std build by creating a new project, then adding a `.cargo/config` as appropriate, something like: 
```
[target.thumbv7em-none-eabihf]
rustflags = [
   "-C", "linker=arm-none-eabi-gcc",
   "-C", "link-arg=-Wl,-Tlink.x",
   "-C", "link-arg=-nostartfiles",
]

[build]
target = "thumbv7em-none-eabihf"
```

You do generally need to set the linker, so it's not quite as easy as just adding a `--target` argument to `cargo`. Cargo generate can also set up a basic no-std project, you don't need to use this but, it's easy to
`cargo generate --git https://github.com/rust-embedded/cortex-m-quickstart -n sodiumoxide-nostd`

Then add `sodiumoxide` and try to actually build takes something with something link (modified to suit this PR):
```
SODIUM_DISABLE_PIE=1 LDFLAGS='--specs=nosys.specs' CLAGS='-DLIBSODIUM_NO_PLATFORM_RANDOM=1' SODIUM_HOST=arm-none-eabi cargo build
```

The targets that are don't work are the `thumbv*` ones from `rustup target list | grep thumb`, ideally any fixes should solve this for _all_ of em rather than being hardcoded for one.

adamreichold:
@ryankurte Could you update the libsodium submodule in a separate PR as this also requires updating binary files stored in the repository.

ryankurte:
> @ryankurte Could you update the libsodium submodule in a separate PR as this also requires updating binary files stored in the repository.

done! not sure why that particular test has failed on `1.36.0` as it worked on every previous actions run :-/

src/version.rs
```diff
@@ -1,21 +1,31 @@
...
 
 use ffi;
-use libc;
+
 use std::slice;
 use std::str;
 
 /// `version_string()` returns the version string from libsodium.
 pub fn version_string() -> &'static str {
     // Use libc directly because CStr isn't available with #![no_std] :(
```
kpp, line 10: fix comment

.github/workflows/default.yml
```diff
@@ -91,7 +91,7 @@ jobs:
     - uses: actions-rs/cargo@v1
       with:
         command: test
-        args: --no-default-features
+        args: --no-default-features --features=cty
```
kpp, line 94: Does it mean we have to enable feature cty in order to compile in no_std env?

ryankurte: technically it's orthogonal to the `libc` feature, where `libc` has the appropriate types for the platform you could use that, where it doesn't you'll need `cty`

kpp: Do we need libc at all (except for tests in src/utils.rs)?

ryankurte: possibly not? but it's also normal to use libc when std is available

adamreichold: Personally, I would also prefer a simpler dependency structure, e.g. always use `cty` instead of `libc` and have `libc` only as a dev-dependency if it is required for test code.

kpp: > Personally, I would also prefer a simpler dependency structure, e.g. always use cty instead of libc and have libc only as a dev-dependency if it is required for test code.

Yes

libsodium-sys/Cargo.toml
```diff
@@ -15,17 +15,16 @@ version = "0.2.6"
 pkg-config = "0.3"
 walkdir = "2"
 
-[target.'cfg(target_env = "msvc")'.build-dependencies]
-libc = { version = "0.2" , default-features = false }
```
wusyong, line 19: This needs to be replaced by `cty` too.

dnaq:

Issue #430, opened 2020-07-23
Expose randombytes_buf_deterministic - very useful for writing tests


tasn:
https://libsodium.gitbook.io/doc/generating_random_data

tasn:
Thanks a lot!

Issue #426, opened 2020-07-20
Support the sodium padding utility functions


tasn:
These ones: https://libsodium.gitbook.io/doc/padding

Thanks!

PR #431, opened 2020-07-29
feat: expose randombytes_buf_deterministic

klittlepage:
Exposes randombytes_buf_deterministic and addresses
closes https://github.com/sodiumoxide/sodiumoxide/issues/430

### DIFF
--- a/src/randombytes.rs
+++ b/src/randombytes.rs
@@ -4,6 +4,11 @@ use ffi;
 #[cfg(not(feature = "std"))]
 use prelude::*;
 
+/// The number of seed bytes to use for the deterministic RNG functions
+/// [`randombytes_buf_deterministic()`] and
+/// [`randombytes_buf_deterministic_into()`]
+pub const SEEDBYTES: usize = ffi::randombytes_SEEDBYTES as usize;
+
 /// `randombytes()` randomly generates size bytes of data.
 ///
 /// THREAD SAFETY: `randombytes()` is thread-safe provided that you have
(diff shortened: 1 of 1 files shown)


src/randombytes.rs
```diff
@@ -1,9 +1,16 @@
 //! Cryptographic random number generation.
 
+#![warn(intra_doc_link_resolution_failure)]
```
jyn514, line 3: This is on by default in nightly, you don't need to add it explicitly.

src/randombytes.rs
```diff
@@ -1,9 +1,16 @@
...
+
 use ffi;
 #[cfg(not(feature = "std"))]
 use prelude::*;
 
+/// The number of seed bytes to use for the deterministic RNG functions
+/// [randombytes_uniform] and
+/// [randombytes_buf_deterministic_into]
```
jyn514, line 11:
```suggestion
/// [`randombytes_uniform()`] and
/// [`randombytes_buf_deterministic_into()`]
```

src/randombytes.rs
```diff
@@ -41,6 +48,35 @@ pub fn randombytes_uniform(upper_bound: u32) -> u32 {
...
+new_type! {
+    /// `Seed` bytes for the deterministic random functions
+    secret Seed(SEEDBYTES);
+}
+
+/// The `randombytes_buf_deterministic` function stores size bytes into buf indistinguishable from
+/// random bytes without knowing seed. For a given seed, this function will always output the same
+/// sequence; size can be up to 2^38 (256 GB).
+///
+/// Seed is [SEEDBYTES] bytes long.
```
jyn514, line 60:
```suggestion
/// Seed is [`SEEDBYTES`] bytes long.
```

src/randombytes.rs
```diff
@@ -41,6 +48,35 @@ pub fn randombytes_uniform(upper_bound: u32) -> u32 {
...
+/// the hood, it uses the ChaCha20 stream cipher. Up to 256 GB can be produced with a single seed.
+pub fn randombytes_buf_deterministic(size: usize, seed: &Seed) -> Vec<u8> {
+    unsafe {
+        let mut buf = vec![0u8; size];
+        ffi::randombytes_buf_deterministic(buf.as_mut_ptr() as *mut _, size, seed.0.as_ptr());
+        buf
+    }
+}
+
+/// A counterpart to [randombytes_buf_deterministic] that
```
jyn514, line 72:
```suggestion
/// A counterpart to [`randombytes_buf_deterministic()`] that
```

src/randombytes.rs
klittlepage: 👍

src/randombytes.rs
klittlepage: 👍

src/randombytes.rs
klittlepage: 👍

src/randombytes.rs
```diff
@@ -41,6 +46,35 @@ pub fn randombytes_uniform(upper_bound: u32) -> u32 {
     unsafe { ffi::randombytes_uniform(upper_bound) }
 }
 
+new_type! {
+    /// `Seed` bytes for the deterministic random functions
+    secret Seed(SEEDBYTES);
```
kpp, line 51: Why secret?

dnaq: It has to be if you want to use it to generate keys and such.

klittlepage: I debated using a `Nonce` type but semantically it is a secret given that the RNG is purely deterministic given knowledge of the initial seed.

kpp: Why not “pk”?

klittlepage: I'm not sure that I follow. Do you mean `secret` vs `public` in `new_type` or do you mean as a naming convention?

It's closer to a secret than it is to a public key IMO. As the libsodium docs state:

> the randombytes_buf_deterministic function stores size bytes into buf indistinguishable from random bytes without knowing seed.

> For a given seed, this function will always output the same sequence. size can be up to 2^38 (256 GB).

It goes on to state that:

> This function is mainly useful for writing tests, and was introduced in libsodium 1.0.12

but using a PRNG with a secret seed and relying on the seed remaining secret isn't unheard of.

kpp: I see.

src/randombytes.rs
```diff
@@ -41,6 +46,35 @@ pub fn randombytes_uniform(upper_bound: u32) -> u32 {
...
+}
+
+/// The `randombytes_buf_deterministic` function stores size bytes into buf indistinguishable from
+/// random bytes without knowing seed. For a given seed, this function will always output the same
```
dnaq, line 55: This needs to be documented better. Add a more generic description of what randombytes_deterministic is, when it should be used, that it should not be used outside of testing and so on. You could probably use a lot of the documentation from libsodium when applicable.

We don't really ever want users to use these functions outside of testing since the whole security of their crypto system will end up depending on the seed.

klittlepage: I pushed updated docs with a warning up top. Note that the docs from the earlier push are 1:1 with libsodium.

klittlepage:
> Maybe add some sanity tests that make sure that different seeds generate different random sequences.

@dnaq done

dnaq:
Great. Just rebase and ping @kpp for merging.

klittlepage:
@kpp rebased

@dnaq 

https://github.com/sodiumoxide/sodiumoxide/pull/418 is approved/ready to merge and https://github.com/sodiumoxide/sodiumoxide/pull/432 got opened yesterday to close https://github.com/sodiumoxide/sodiumoxide/issues/426. I know that you just did a release last week but in the course of merging this is there any chance that I could submit a change log and a version bump to get these other PRs merged and released? Doing so would resolve three issues that other users had opened and selfishly I'd love to do stop relying on my own fork.

Short of a release just getting these PRs onto master would be helpful.

kpp:
bors: r+

PR #434, opened 2020-09-11
Use `ed25519::Signature` type; impl `Signer`/`Verifier` traits

tarcieri:
The `ed25519` crate provides a common `Signature` type which can be shared among libraries which implement Ed25519 signing/verification:

https://docs.rs/ed25519

Without common types/traits for Ed25519 signatures, downstream consumer crates which use Ed25519 as part of a high-level protocol implementation which allows consumers of their crates to be able to use their choice of Rust Ed25519 libraries wind up writing their own abstractions to do so on a crate-by-crate basis.

The `ed25519` crate, and the `signature` crate it's built on, provide a common abstraction layer:

- `ed25519` crate: https://github.com/RustCrypto/signatures/tree/master/ed25519
- `signature` crate: https://github.com/RustCrypto/traits/tree/master/signature

The `ed25519` crate has been natively adopted by `ed25519-dalek`:

- Issue: https://github.com/dalek-cryptography/ed25519-dalek/issues/80
- Usage example: https://docs.rs/ed25519/1.0.2/ed25519/#using-above-example-with-ed25519-dalek

If `sodiumoxide` were to adopt these types/traits as well, I'd be happy to add similar documentation to the `ed25519` crate for how to use them with `sodiumoxide`.

### DIFF
--- a/src/crypto/sign/ed25519.rs
+++ b/src/crypto/sign/ed25519.rs
@@ -3,6 +3,11 @@
 //! standard notion of unforgeability for a public-key signature scheme under
 //! chosen-message attacks.
 
+pub use ed25519::{
+    signature::{Signer, Verifier},
+    Error, Signature,
+};
+
 use ffi;
 use libc::c_ulonglong;
 #[cfg(not(feature = "std"))]
@@ -52,14 +57,25 @@ impl SecretKey {
     }
 }
 
+impl Signer<Signature> for SecretKey {
+    fn try_sign(&self, msg: &[u8]) -> Result<Signature, Error> {
+        Ok(sign_detached(msg, self))
+    }
+}
+
 new_type! {
     /// `PublicKey` for signatures
     public PublicKey(PUBLICKEYBYTES);
 }
 
-new_type! {
-    /// Detached signature
-    public Signature(SIGNATUREBYTES);
+impl Verifier<Signature> for PublicKey {
+    fn verify(&self, msg: &[u8], sig: &Signature) -> Result<(), Error> {
+        if verify_detached(sig, msg, self) {
+            Ok(())
+        } else {
+            Err(Error::new())
+        }
+    }
 }
 
 /// `gen_keypair()` randomly generates a secret key and a corresponding public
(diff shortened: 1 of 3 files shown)


src/crypto/sign/ed25519.rs
```diff
@@ -389,20 +402,21 @@ mod test {
...
+    //     use test_utils::round_trip;
+    //     for i in 0..256usize {
+    //         let (pk, sk) = gen_keypair();
+    //         let m = randombytes(i);
+    //         let sig = sign_detached(&m, &sk);
+    //         round_trip(pk);
+    //         round_trip(sk);
+    //         round_trip(sig);
+    //     }
+    // }
```
tarcieri, line 419: Note: the `ed25519` crate has a `serde` feature which provides 64-byte serialized signatures in formats like bincode and CBOR.

I didn't wire it up yet in this PR since enabling a `serde` feature in a dependency when `serde` is already an optional crate dependency is a bit tricky. That said, we can probably use the same workaround I used in ed25519-dalek:

https://github.com/dalek-cryptography/ed25519-dalek/blob/952bdd0/Cargo.toml#L30

tarcieri:
FYI, I already maintain a "wrapper crate" which enables the usage of these Ed25519 signature types/traits:

https://docs.rs/signatory-sodiumoxide/0.21.0/signatory_sodiumoxide/

However, I think it'd be great if `sodiumoxide` added native support the same way `ed25519-dalek` did.

tarcieri:
Note: per #445 I'm definitely still interested in seeing this through. Just wanted to make sure there's actually interest in this PR before completing it.

kpp:
Yes I am interested in this PR.

tarcieri:
The tricky remaining question is what to do about `serde` support. If you'd like to use the built-in support in the `ed25519` crate, it'd require using a hack similar to the one in `ed25519-dalek`:

https://github.com/dalek-cryptography/ed25519-dalek/pull/124/files#r393865319

If that seems acceptable I can do the same thing, but I just wanted to make sure first.

Note that there's been quite a bit of discussion about `serde` support in `ed25519-dalek`, and it seems like the serializers the `ed25519` crate provides may differ from the ones you presently define. See this issue for some additional discussion:

https://github.com/dalek-cryptography/ed25519-dalek/pull/146

kpp:
TBO I personally can't find `serde` feature useful at all... However we can work on it.

tarcieri:
Okay, I can update the PR to do what I did in ed25519-dalek and we can go from there.

tarcieri:
I went ahead and rebased this PR, and for now just removed the `serde` test, as well as WIP.

I think it's pretty problematic to serialize Ed25519 signatures using `serde`. See the discussion on this issue:

https://github.com/dalek-cryptography/ed25519-dalek/pull/146

Without looping in an additional crate like `serde_bytes`, the serialization will be bad for at least one of: bincode, CBOR, JSON.

So I left it out for now, but if you'd like me to revisit that I can.

kpp:
@sodiumoxide/owners  How about breaking `serde` feature and revisiting in sometimes later when this feature is requested?

kpp (approved):

dnaq:
Agreed, as it stands now serde support is a misfeature and it will keep
being one until the derive-macros work on large arrays, but that will break
the serialization format anyway.

kpp:
bors: r+

PR #435, opened 2020-10-07
Enable mingw32 build

cjdelisle:
This patch makes it possible to build for windows using mingw32

### DIFF
--- a/libsodium-sys/build.rs
+++ b/libsodium-sys/build.rs
@@ -199,6 +199,10 @@ fn make_libsodium(target: &str, source_dir: &Path, install_dir: &Path) -> PathBu
             _ => panic!("Unknown iOS build target: {}", target),
         }
         cross_compiling = true;
+    } else if target.eq("i686-pc-windows-gnu") {
+        host_arg = "--host=i686-w64-mingw32".to_string();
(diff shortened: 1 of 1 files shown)


kpp:
@cjdelisle seems we test only msvc in https://github.com/sodiumoxide/sodiumoxide/blob/master/.github/workflows/default.yml. Would you please add another toolchain to test you changes with mingw32?

laumann:
FWIW we could really use this, and this change appears to work (it builds at least).

libsodium-sys/build.rs
```diff
@@ -216,6 +216,10 @@ fn make_libsodium(target: &str, source_dir: &Path, install_dir: &Path) -> PathBu
             _ => panic!("Unknown iOS build target: {}", target),
         }
         cross_compiling = true;
+    } else if target.eq("i686-pc-windows-gnu") {
+        host_arg = "--host=i686-w64-mingw32".to_string();
+    } else if target.eq("x86_64-pc-windows-gnu") {
+        host_arg = "--host=x86_64-w64-mingw32".to_string();
```
laumann, line 222: Just above the flag `cross_compiling` is set to `true`. Should that flag also be set here? (It doesn't seem to be necessary, just wondering.)

PR #444, opened 2020-11-24
Add cross compile checks to github actions

### LINKED ISSUE #313: Add armv7-unknown-linux-musleablhf to the CI process
Thanks. This is a good start. But we need to add the architecture to the CI process, so we know that the crate works on it and doesn't break in future.

_Originally posted by @Dylan-DPC in https://github.com/sodiumoxide/sodiumoxide/pull/294#issuecomment-443796338_

kpcyrd:
Resolves #313. This doesn't use/test the instructions in our readme though but instead only tests our code works on armv7 (and all other debian architectures). The x86_64 one is musl intentionally so we don't only test gnu libc.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -76,6 +76,37 @@ jobs:
           command: clippy
           args: --all --lib --tests -- --deny warnings
 
+  cross:
+    name: cross
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


kpp (approved):

kpp:
bors: r+

PR #449, opened 2020-12-09
use system-deps to handle libsodium dependency

### LINKED ISSUE #359: unify environment variables with those used in pkg-config

gdesmott:
Breaking change as some old build env variables are now longer working:

- re-implemented SODIUM_LIB_DIR assuming you want to keep it but only
tested on Linux.
- removed SODIUM_SHARED as there is currently no way to statically link
to SODIUM_LIB_DIR. I can consider extending system-deps if you'd like to
keep this feature.

Fix #359


kpcyrd:
Our CI currently relies on static linking, you could add the required environment variables to enable that behavior to `.github/workflows/default.yml`, and you need to follow the instructions [here](https://github.com/rust-embedded/cross#passing-environment-variables-into-the-build-environment) so it's correctly passing them through to cross.

Reach out if you need help and I'd try to prepare a patch to do that. :)

gdesmott:
Thanks @kpcyrd 

I defined the proper env variable but I'm not sure about where to put this `Cross.toml` file?

gdesmott:
Also, looks like Windows building is failing because it doesn't find the pkg-config file. Isn't it shipped with the windows lib?

kpp:
Yeah our build.rs is a bit complicated. On linux we can take a system lib or build libsodium from source (if crosscompiling), on windows we can either link via a provided lib or take it from https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/build.rs#L324.

kpp:
@gdesmott  Also take a look at https://github.com/sodiumoxide/sodiumoxide/issues/377

kpp:
Nice progress. Now we are left with env vars https://github.com/sodiumoxide/sodiumoxide/blob/master/libsodium-sys/README.md (tested in https://github.com/sodiumoxide/sodiumoxide/blob/master/testcrate/build.rs).

gdesmott:
Yeah I'm looking into it, it's a bit trickier than expected.

kpp:
Need help?

gdesmott:
Didn't had much time to investigate lately because of work.

The problem is with `DEP_SODIUM_LIB`. system-deps [follow the pkg-config default](https://github.com/gdesmott/system-deps/commit/07b8b703ff448193b9922a63d423794f6b549b7a) and so does not return the lib dir when using the standard location. See [this discussion](https://github.com/rust-lang/pkg-config-rs/pull/35) for the rational of this change.

As a result I can't guarantee to have the libdir defined from pkg-config which makes things harder. We could of course re-pull the lib with `PKG_CONFIG_ALLOW_SYSTEM_LIBS` but it kinda defeat the purpose of using  `system-deps`. :\

gdesmott:
I added some new API to system-deps to try solving this problem. It's not merged/released yet as I'd like to make sure it does the job before.

Looks like we have two issues left:
- 1.36 support. Would you consider bumping the minimum Rust version or should I try solving this in system-deps?
- Windows test keep failing, I'm not sure why. I don't have the setup to reproduce it locally so some help would be welcome here @kpp

kpp:
> Would you consider bumping the minimum Rust version

Yes

dnaq:

Issue #221, opened 2018-06-11
Custom Error Types


Dylan-DPC:
Currently in the crate we are doing `Err(())`. However, this isn't verbose enough for the user and he wouldn't know what error has occurred without going through the crate source. Instead we can create custom error types for different types of errors which along with a message, gives a clear meaning to the user about the error. 

A general practice is to use the [failure](https://github.com/rust-lang-nursery/failure) crate.

nvesely:
> A general practice is to use the failure crate.

I've never used this crate, but just skimmed the README to look for no_std compatibility and saw

> failure is no_std compatible, though some aspects of it (primarily the Error type) will not be available in no_std mode.

So we'd have to do something like

```rust
    if cfg!(feature = "std") {
        return failure::Error();
    } else {
        return Err(());
    }
```

universally. Depending on what type of error-handling we want to implement, this could become more complicated.

nvesely:
Wait that's not going to work. We'll have to write each function twice because the return types are different. Maybe this means we'll want to go a simpler route for error-handling.

Dylan-DPC:
@nvesely ah yes forgot about the no_std part. We will keep this pending for a while till we find a good solution for this

dnaq:
Can't we just create our own custom error type? Users of our crate can wrap it in failure if needed or do whatever they want with it.

naftulikay:
Yes, totally agree with @dnaq

nvesely:
Third for keeping it simple. Don't think we need to pull in additional crates for this one.

Dylan-DPC:
I agree we may not need a crate right now for this. But failure is there for certain reasons, so in future there is a good chance we may be needing it. Can skip for now though.

caass:
Hello! Is there any ongoing work on this issue? If not, I'd like to give it a try.

For an example use case, I'm trying to use this library with [juniper](https://github.com/graphql-rust/juniper), which will perform [automatic error conversion](https://graphql-rust.github.io/juniper/master/types/objects/error_handling.html#field-results) with anything that implements `std::fmt::Display`.

Creating a custom error type that implements `Display` would make my code a lot more ergonomic!

mso42:
Created a PR with a sketch of a pattern for custom error types.

PR #454, opened 2021-01-10
WIP: Custom error type for secretbox::open and secretbox::open_detached

mso42:
Adds a custom error type `SecretBoxError`. Is this pattern good enough to implement everywhere?

Addresses https://github.com/sodiumoxide/sodiumoxide/issues/221

### DIFF
--- a/src/crypto/box_/curve25519xsalsa20poly1305.rs
+++ b/src/crypto/box_/curve25519xsalsa20poly1305.rs
@@ -246,7 +246,6 @@ pub fn seal_precomputed(m: &[u8], n: &Nonce, k: &PrecomputedKey) -> Vec<u8> {
     let clen = m.len() + MACBYTES;
     let mut c = Vec::with_capacity(clen);
     unsafe {
-        c.set_len(clen);
         ffi::crypto_box_easy_afternm(
             c.as_mut_ptr(),
             m.as_ptr(),
@@ -254,6 +253,7 @@ pub fn seal_precomputed(m: &[u8], n: &Nonce, k: &PrecomputedKey) -> Vec<u8> {
             n.0.as_ptr(),
             k.0.as_ptr(),
         );
+        c.set_len(clen);
     }
     c
 }
@@ -285,20 +285,20 @@ pub fn open_precomputed(c: &[u8], n: &Nonce, k: &PrecomputedKey) -> Result<Vec<u
     }
     let mlen = c.len() - MACBYTES;
     let mut m = Vec::with_capacity(mlen);
-    let ret = unsafe {
-        m.set_len(mlen);
-        ffi::crypto_box_open_easy_afternm(
+    unsafe {
+        let ret = ffi::crypto_box_open_easy_afternm(
             m.as_mut_ptr(),
             c.as_ptr(),
             c.len() as u64,
             n.0.as_ptr(),
             k.0.as_ptr(),
-        )
-    };
-    if ret == 0 {
-        Ok(m)
-    } else {
-        Err(())
+        );
+        if ret == 0 {
+            m.set_len(mlen);
+            Ok(m)
+        } else {
+            Err(())
+        }
     }
 }
 
--- a/src/crypto/secretbox/xsalsa20poly1305.rs
+++ b/src/crypto/secretbox/xsalsa20poly1305.rs
@@ -68,7 +68,6 @@ pub fn seal(m: &[u8], n: &Nonce, k: &Key) -> Vec<u8> {
     let clen = m.len() + MACBYTES;
     let mut c = Vec::with_capacity(clen);
     unsafe {
-        c.set_len(clen);
         ffi::crypto_secretbox_easy(
             c.as_mut_ptr(),
             m.as_ptr(),
@@ -76,6 +75,7 @@ pub fn seal(m: &[u8], n: &Nonce, k: &Key) -> Vec<u8> {
             n.0.as_ptr(),
             k.0.as_ptr(),
         );
+        c.set_len(clen);
     }
     c
 }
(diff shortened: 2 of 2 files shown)


src/crypto/secretbox/error.rs
```diff
@@ -0,0 +1,25 @@
...
+#[cfg(feature = "std")]
+impl std::fmt::Display for SecretBoxError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for SecretBoxError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
```
mexus, line 22: It's a default implementation, isn't it? https://doc.rust-lang.org/src/std/error.rs.html#100 So the whole block is equivalent to `impl std::error::Error for SecretBoxError {}`

src/crypto/secretbox/error.rs
```diff
@@ -0,0 +1,25 @@
...
+    /// function, which expects a message prepended with a MAC.
+    CiphertextTooShort,
+    /// The ciphertext failed verification.
+    CiphertextFailedVerification,
+}
+
+#[cfg(feature = "std")]
+impl std::fmt::Display for SecretBoxError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
```
mexus, line 16: Normally, the `Display` trait is used to provide human-readable error messages, target at end users. Won't it make better sense to display what is currently written in doc comments instead?

src/crypto/secretbox/error.rs
```diff
@@ -0,0 +1,25 @@
...
+#[derive(Debug, PartialEq)]
+pub enum SecretBoxError {
+    /// The ciphertext was too short. This results from using the `crypto_secretbox_open_easy`
+    /// function, which expects a message prepended with a MAC.
+    CiphertextTooShort,
+    /// The ciphertext failed verification.
+    CiphertextFailedVerification,
+}
+
+#[cfg(feature = "std")]
```
mexus, line 13: Since `std::fmt::Display` is an alias for `core::fmt::Display`, it is possible to provide better user experience for no-std users by implementing the latter trait unconditionally, even though the `Error` trait could not be implemented with absence of std

mso42:
All good feedback, thanks

src/crypto/secretbox/error.rs
mso42: Agreed. Do you know of an easy way to do this wthat doesn't involve duplicating the text in the doc comments?

mexus: The only way I'm aware of is to use a custom macro :) like that one: [displaydoc](https://crates.io/crates/displaydoc) crate

mso42:
New commit resolving the above comments and using `displaydoc` crate to derive Display for the error.

Adding a dependency is somewhat controversial I guess, but:
- it has minimal child dependencies (standard macro stuff like `proc_macro2`, `syn`, `quote`)
- it works in `no-std`

The alternative is to duplicate the display strings in the documentation comments and the `Display` implementation.

kpp:
> New commit resolving the above comments and using displaydoc crate to derive Display for the error.

Looks like we are not far away from snafu :)

dnaq:
Yeah, I feel like a whole new dependency for not having to write a display
implementation is a bit overkill.

mexus:
> > New commit resolving the above comments and using displaydoc crate to derive Display for the error.
> 
> Looks like we are not far away from snafu :)

snafu supports no_std btw :) and with its default features turned off it introduces only syn/quote/proc-macro2 and [doc-comment macro](https://docs.rs/doc-comment) crates as additional dependencies.

> Yeah, I feel like a whole new dependency for not having to write a display implementation is a bit overkill.

It definitely looks like an overkill when writing a display implementation only once or twice, but what will it look like when more error handling to be introduced?

mso42:
I think that including the entire docstring as the Display message for the error may not always be the right thing to do. For example, if the docstring (for some reason) were multiple lines with code snippets etc. etc.

My preference is to keep error messages reasonably short, so that they can be pasted into search engines and relevant documentation found from there. This is why I simply used the Debug string for the Display in the initial patch.

I note that in our case, our error types are probably not going to be that complicated. (Probably just enums with no fields.) And in that case, I think something like snafu or thiserror is even more overkill than displaydoc.

mso42:
I have got rid of the display-doc stuff, and written some error messages that say nothing more than the enum field, but in plain English.

Also squashed all the commits so far.

mso42:
While we are still pending agreement on the actual text of the error messages, can I get feedback on the general approach here, i.e. having one error type per type of operation?

Or would it be better to have one big `SodiumOxideError` type at the crate root and have everything which could error return instances of that.

My preference is for the former, so that handling code doesn't have to handle `pwhash`-related errors if it is calling `sealedbox` functions (for example).

The argument in favour of the latter is that it keeps things simpler within the crate I guess.

dnaq:
One error type at the crate root. Much better usability.

mso42:
So calling code would have to do:

```rust
match secretbox::seal_detached(&mut buf, &nonce, &key) {
    Ok(tag) => { /* ... */ },
    Err(SodiumError::SecretBoxSealError) => { /* handle error */ },
    Err(_) => { /* how could this possibly even happen? what do we do here? panic? */ },
}
```

I don't see how that is better usability.

dnaq:
For any library/binary that doesn’t want to handle the error it becomes
something like:

```
enum Error {
  Foo(...),
  Bar(...),
  CryptoError(sodiumoxide::Error),
}
```

mso42:
Oh yes that is true. What do other libraries do?

dnaq:
Most that I’ve seen, reqwest is one.

adamreichold:
> So calling code would have to do:

Another thing to keep in mind is that while SNAFU-style nested error enums can allow exhaustive matching, this can imply problems with backwards compatibility. Meaning we might want to make all our error enums `#[non_exhaustive]` so that even the specialised enums could not be exhaustively and would always need a "unexpected fatal error" match arm.

kpp:
@mso42 why did you close this PR?..

mso42:
Oh, I didn't mean to. I renamed my working branch and reset my repository's `master` to match upstream. Looks like that auto-closed the PR.

I have done a little work on this, and intend to do more, but it is a big job and I'm not sure I have the motivation to see it through. If I get to something that I can believably call not a work-in-progress, I'll reopen another PR.

PR #455, opened 2021-01-10
Do not call Vec::set_len before the data is initialised

adamreichold:
While this should not be exploitable as `u8` is `Copy` and hence is not dropped, it seems prudent to only change the length of a `Vec` after the data has been initialised via an FFI call (as is already done in the other places that use `Vec::set_len`).


kpp (approved):

kpp:
bors: r+

PR #461, opened 2021-05-31
Rebuild documentation for Github pages

tomstoneham:
Thanks so much for creating sodiumoxide! I use it all the time :)

Currently the documentation available at https://sodiumoxide.github.io/sodiumoxide is quite out of date. It was last built 3 years ago, and doesn't include the `sodiumoxide::crypto::secretstream` API for example. This commit rebuilds the documentation as of the current `master` branch, the result can be seen at https://tomstoneham.github.io/sodiumoxide

Honestly, the better option might just be to delete the `gh-pages` branch entirely: The documentation at https://docs.rs/sodiumoxide should always be up-to-date with the current release on crates.io. This commit is just for if you would prefer to keep the github pages site, I think it should at least be up to date.


dnaq:
Deleted the branch. Thanks for bringing the issue to our attention.

tomstoneham:
Thanks so much for the quick response! Have a good one.

PR #463, opened 2021-05-31
Document crypto::generichash

tomstoneham:
This PR adds documentation for the `crypto::generichash` module, indicating its intended usage, and linking to other parts of the crate which are better suited for various purposes. It also exposes a couple of constants from libsodium which define the recommended digest size/key size for generichash, and removes a dead link from the `libsodium-sys` Cargo.toml.


kpcyrd (approved):

kpcyrd:
I've reviewed the PR and it looks good, it seems to need a rebase to be merged though.

Thanks!

PR #464, opened 2021-05-31
crypto::secretstream::Stream: Implement push/pull_to_slice methods

tomstoneham:
Hi! This PR introduces two new methods to the public API for [`crypto::secretstream::Stream`](https://docs.rs/sodiumoxide/0.2.6/sodiumoxide/crypto/secretstream/index.html), as well as moving around some of the internals.

The two new methods are `Stream<Push>.push_to_slice()` and `Stream<Pull>.pull_to_slice()`. These work much like the similar `push_to_vec` and `pull_to_vec` methods, except they accept an `&mut [u8]` rather than an `&mut Vec<u8>`, and it is this slice to which the resulting ciphertext/plaintext is read. The motivation for this change is to allow users to directly read messages from a secretstream into their own buffer types. For example, if a user is decrypting a plaintext that they want to zero after use, they might prefer to write directly to a buffer implementing [`Zeroize`](https://docs.rs/zeroize/1.3.0/zeroize/), or using a hardened memory region allocated with [`sodium_malloc`](https://doc.libsodium.org/memory_management#guarded-heap-allocations), rather than first writing the plaintext to a vector.

These methods check at runtime that the slice is of sufficient length, and return an `Err` if it is not.

Internally, the PR creates new private methods `push_impl_ptr` and `pull_impl_ptr`, which are very similar to `push_impl` and `pull_impl`, but which take a mutable pointer directly, rather than taking a vector and then using `as_mut_ptr` to attain a mutable pointer. Otherwise the functionality is the same as the previous implementations. `push_impl` and `pull_impl` are rewritten to rely on these new methods, but can (hopefully!) be fairly simply seen to have identical functionality.

This is one of my first larger PRs to an external repo, so I would greatly appreciate any feedback :) Cheers!

### DIFF
--- a/src/crypto/secretstream/secretstream_macros.rs
+++ b/src/crypto/secretstream/secretstream_macros.rs
@@ -226,6 +226,26 @@ impl Stream<Push> {
         }
     }
 
+    /// All data (including optional fields) is authenticated. Encrypts a
+    /// message `m` and its `tag`. Optionally includes additional data `ad`,
+    /// which is not encrypted.
+    ///
+    /// The encrypted message is written to the `out` slice, which must be at least `m.len()` +
+    /// [`ABYTES`](crate::crypto::secretstream::ABYTES) long, overwriting the first `m.len() +
+    /// ABYTES` bytes. An error will be returned if the slice is not long enough. If the push is
+    /// successful, the length of data written to the slice will be returned.
+    pub fn push_to_slice(&mut self, m: &[u8], ad: Option<&[u8]>, tag: Tag, out: &mut [u8]) -> Result<usize, ()> {
+        let buf_len = self.push_check(m, tag)?;
+        if out.len() < buf_len {
+            return Err(());
+        }
+        // SAFETY: The previous if block ensures that the slice has at least m.len() + $abytes of
+        // capacity.
+        unsafe {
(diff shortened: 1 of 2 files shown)


src/crypto/secretstream/secretstream_macros.rs
```diff
@@ -232,28 +260,35 @@ impl Stream<Push> {
         Ok(m_len + ABYTES)
     }
 
-    fn push_impl(&mut self, m: &[u8], ad: Option<&[u8]>, tag: Tag, buf: &mut Vec<u8>) -> Result<(), ()> {
+    // SAFETY: buf must have a capacity of at least m.len() + $abytes.
+    unsafe fn push_impl(&mut self, m: &[u8], ad: Option<&[u8]>, tag: Tag, buf: &mut Vec<u8>) -> Result<(), ()> {
```
adamreichold, line 264: Instead of making this unsafe, why not move the two `reserve` calls into this and thereby ensure the invariant locally? (Doing `Vec::new` followed by `Vec::reserve` should not be measurably less efficient than `Vec::with_capacity`.)

tomstoneham: Yeah that makes sense, I'll make a commit. Thanks!

dnaq:

